PR #3, opened 2022-11-29
Update indexing.md

Tuangtuang:
The original proto is confusing because we never know where is com.vectara.indexing and the document definition itself is under this page. 

Signed-off-by: Jean.T <tangyuqi00@qq.com>


eskibars:
Closing this out.  Sorry for the delay.  But we want to make sure to fully qualify something as "generic" as "Document", so we've made sure that's fully qualified while other parts of the API are accessible without a full prefix

PR #4, opened 2022-11-29
Add indexing endpoint

Tuangtuang:
When I tried to use RPC call to index my text, I didn't know what is the indexing endpoint. I searched it on the portal and found nothing and then I searched your github source code to get the answer. I think it will be more convenient and user-friendly if you can put the endpoint on the document page.

Signed-off-by: Jean.T <yt2101@nyu.edu>

### DIFF
--- a/www/docs/indexing-apis/indexing.md
+++ b/www/docs/indexing-apis/indexing.md
@@ -17,6 +17,9 @@ to do that.
 
 The full definition of the gRPC interface is covered below.
 
+### Index Endpoint
+The indexing endpoint is `indexing.vectara.io`.
+
 ### Service
 
 The indexing service operates in two modes: _incremental_ and _batch_. In


eskibars:
Thanks for the PR!  Note that we do have documentation on the REST endpoints at https://docs.vectara.com/docs/rest but I can understand that it would be more convenient to be on the pages for the specific APIs.

A few changes I'd like though:
1. I'd like to be consistent and include this on the various API pages (search and admin)
2. We actually have this all in variables so that if we change API formats in the future, we can make all of the changes in 1 place.  The variables are stored at https://github.com/vectara/vectara-docs/blob/main/www/static/variables.json and an example of the usage is at https://github.com/vectara/vectara-docs/blob/main/www/docs/indexing-apis/file_upload.md?plain=1#L13 .  So this PR would become something like "The indexing endpoint is <Config v="domains.rest.indexing"/>" to help future proof it

If you'd like to do these, let me know and feel free.  If you'd prefer that I close out this PR to replace it with my own that does those 2, let me know that

eskibars:
Thanks again for the PR here!  I hadn't heard back, and I know that what I asked would be significantly more work, so I went ahead and did that work myself and rolled it out.  You can now find indexing endpoints at the tops of the following documentation pages:
- https://docs.vectara.com/docs/indexing-apis/indexing
- https://docs.vectara.com/docs/indexing-apis/deleting-documents
- https://docs.vectara.com/docs/admin-apis/create-corpus
- https://docs.vectara.com/docs/admin-apis/delete-corpus
- https://docs.vectara.com/docs/admin-apis/reset-corpus
- https://docs.vectara.com/docs/search-apis/search

I'm going to go ahead and close this PR out, but thanks again for pointing this out and offering to help fix it!

PR #17, opened 2023-01-24
Add OpenAPI specification plugin

eskibars:

PR #18, opened 2023-01-26
Full replacement with fresh docusaurus install.

cjcenizal:
I reimplemented https://github.com/vectara/vectara-docs/pull/17 from a fresh Docusaurus install, which seemed to fix the problem with the playground not loading properly.

## Changes

* The diff is heavily weighted towards deletions because the change in dependencies results in a much smaller `package-lock.json` file.
* Some `description` information was removed from each generated doc, and I'm not sure why. But it doesn't look critical so it's possibly non-blocking?

## Steps to test

Run `yarn build && yarn serve`. This mimics a production server and enables reproduction of the original bug on `main` (though will require adding the `serve` script to `package.json` on `main`).

## Screenshots

image


www/docs/rest-api/file-upload.api.mdx
```diff
@@ -5,7 +5,7 @@ description: "FileUpload"
...
...returns the extracted document that was indexed"}],"description":"FileUpload","method":"post","path":"/v1/upload","servers":[{"url":"https://api.vectara.io"}],"security":[{"oAuth":[]}],"securitySchemes":{"oAuth":{"type":"oauth2","flows":{"clientCredentials":{"tokenUrl":"https://vectara-prod-YOUR_VECTARA_CUSTOMER_ID.auth.us-west-2.amazoncognito.com/oauth2/token"},"scopes":{}}}},"info":{"title":"Vectara REST API","version":"1.0.0","termsOfService":"https://vectara.com/legal/online-customer-agreement/","x-logo":{"url":"https://docs.vectara.com/img/vectara_wordmark.png","altText":"Vectara"}},"postman":{"name":"File Upload","description":{"type":"text/plain"},"url":{"path":["v1","upload"],"host":["{{baseUrl}}"],"query":[{"disabled":false,"description":{"content":"(Required) Customer ID","type":"text/plain"},"key":"c","value":""},{"disabled":false,"description":{"content":"(Required) Corpus ID","type":"text/plain"},"key":"o","value":""},{"disabled":false,"description":{"content":"(Required) If true, the server returns the extracted document that was indexed","type":"text/plain"},"key":"d","value":""}],"variable":[]},"header":[{"key":"Accept","value":"application/json"}],"method":"POST"}}
```
cjcenizal, line 8: For some reason this has been removed from the API definition:

```js
"description":"Vectara is a neural search platform, built for developers to get the most out of their data.\n\nYou can sign up for an account at [https://vectara.com](https://vectara.com)."
```

cjcenizal: @eskibars Is this missing information a blocker?

eskibars: Thanks for this!  No, it's not: I think what happened here was I hadn't checked in the last OpenAPI spec that I had been working on.  I've continued to work on it in the background anyway, so once we get this merged in, I can just update the latest spec and regen any changed pages (probably only the info page right now)

eskibars (approved):
Thanks for the quick turnaround on this!  Approved.  Specifically:
- Don't worry about the indexing test failure.  We aren't using that yet, and I've been using our own docs repo to test, so you've found another bug in the process, which is great.  I'll find/fix it async
- Don't worry about the delta on the info/description either.  There are a few things I'm planning on updating in the OAS file, but can do that after we move forward with this.  No blocker in the interim

PR #22, opened 2023-02-16
latest text and images changes for new api functionality

danielone:
latest text and images changes for new api functionality


danielone:
Ready for review

@aamirbutt @talip please review the key usage part (JWT) and add any changes that need to go there as I did not update that section.

danielone:
@aamirbutt That sounds fine as to  when to tackle changes. 

Just so I am not blocking and to clarify the ask, please review this specific existing section which has not changed to determine what should change in the new version of this API key doc.
https://docs.vectara.com/docs/api-keys#using-a-key

I would recommend to make it clear how to use an api key for query case as well as that case hits both key usage cases:
query api access
query + index api access

> Hi Daniel,
> 
> I think it would be better/easier if you complete this first and merge the changes. After that, I will create a separate PR for the changes related to code. Ideally, we would like an indexing/uploading example for all three languages (JS, Python and cURL)

eskibars:
I'm closing this PR out as it's grown stale.  We can revisit some of the core ideas in it based on more recent branches

PR #28, opened 2023-03-15
A few minor changes

ofermend:
minor typos.
It was somewhat confusing to me that when I clicked on "advanced applications" it went directly to the slackbot without explanation

### DIFF
--- a/www/src/pages/index.js
+++ b/www/src/pages/index.js
@@ -23,7 +23,7 @@ const features = [
     description: (
       <>
         Our intuitive cloud-based API makes it easy to index and query
-        your textual data, making it easy to create <a href="https://github.com/vectara/slackbot">advanced applications</a>  quickly.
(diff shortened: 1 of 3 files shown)


www/src/pages/index.js
```diff
@@ -23,7 +23,7 @@ const features = [
     description: (
       <>
         Our intuitive cloud-based API makes it easy to index and query
-        your textual data, making it easy to create <a href="https://github.com/vectara/slackbot">advanced applications</a>  quickly.
+        your textual data, making it easy to create advanced applications like <a href="https://github.com/vectara/slackbot">out slackbot</a> quickly.
```
eskibars, line 26: I presume this is supposed to say `our Slackbot` instead of `out slackbot`?

eskibars (changes requested):
Thanks @ofermend !  1 requested change here

PR #42, opened 2023-05-01
Update protobuf definitions in docs

eskibars:
Unfortunately, we have a synchronization problem with the docs in trying to keep them up to date with the protobuf definitions: it has to be done manually currently.  This PR brings the protobuf definitions up to date with the latest, which is mostly comprised of additional comments that should be helpful to a reader as well, but also adds a few parameters that have been newly minted since we first published this and removes some extraneous/old parameters as well


amin2718 (approved):
Looks good. I gave it a quick look since it's copying over code from platform/

PR #48, opened 2023-05-16
Update OAS spec, add highlighting cfg and lambda to playground

eskibars:
As the title says:
- Just updated to latest protos
- Generated new OAS spec based on those protos
- Regenerated the API playground docs as well


www/static/vectara-oas.yaml
```diff
@@ -411,6 +411,31 @@ components:
          - DEFAULT: Use corpus-assigned semantics.  This is the most common setting.
          - QUERY: Use query semantics.  This is also common.
          - RESPONSE: Use response semantics.  Usage of this is rare.
+    QueryRequestContextConfig:
+      type: object
+      properties:
+        charsBefore:
+          type: integer
+          format: int32
+          description: The amount of context before. Ignored if sentences_before is set.
```
ofermend, line 420: Is it possible to explain these (charsBefore, charsAfter, sentencesBefore, sentencesAfter) a bit more clearly? We assume here that "context" is clear to the reader and it may not always be. Also what is the definition of a "sentence" in this context? The tags are also not entirely clear without further explanation.

www/static/vectara-oas.yaml
```diff
@@ -930,7 +967,22 @@ components:
...
+      properties:
+        lambda:
+          type: number
+          format: float
+          title: >-
+            Controls the weight given to lexical matches. The final score, S, is then:
+
+
+
+                S = (lambda) * (lexical score) + (1 - lambda) * (dense score)
```
ofermend, line 984: Do we want to provide any specific default numbers to use as general guidance for the value of lambda?

www/static/vectara-oas.yaml
```diff
@@ -1004,6 +1098,40 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/servingAttribute"
+    servingSummarizationRequest:
+      type: object
+      properties:
+        summarizerPromptId:
```
ofermend, line 1104: I understand that summarizer and prompt are both included in this Id, but from a user point of view - is it true that we don't really expose what the prompts are? If so, can we find a different way to describe this?

eskibars: Let's take this conversation offline

eskibars:
Closing this PR as it's been superseded

PR #55, opened 2023-05-25
Code samples edit link

### LINKED ISSUE #50: "Edit this page" link from REST examples pages leads to a 404
### Pages

All REST examples pages

### Description

When you select "Edit this page" from any of the noted pages, you get a 404 error. Example:

From https://docs.vectara.com/docs/getting-started-samples/RestUploadFile.java, I scroll down, select "Edit this page", I'm redirected to https://github.com/vectara/vectara-docs/tree/main/www/docs/getting-started-samples/RestUploadFile.java.md

This is an interesting problem, as anyone who wants to edit this page would have to understand that it pulls its source from two other pages.

eskibars:
Fixes #50 by using the `custom_edit_url` which I just found out about.  Thank you for the report @mjang !


eskibars:
Confirmed this works (and has no undesired effects)

PR #61, opened 2023-06-09
Fix typos for "added" and "called"

pwoznic:
The words "added" and "called" were mispelled so I corrected them in the Custom Dimensions topic

### DIFF
--- a/www/docs/common-use-cases/semantic-search/custom_dimensions.md
+++ b/www/docs/common-use-cases/semantic-search/custom_dimensions.md
@@ -95,10 +95,10 @@ END
 In order to calculate the final score of a document and query that contains
 custom dimensions, <Config v="names.product"/> takes the dot product of the
 query's custom dimensions with the document's custom dimensions and the resulting
-number is aded to their score.
+number is added to their score.
 
-Negative values decrease the overall score (sometimes caled "burying") and
-postive values increase the overal score (sometimes caled "boosting").  A dot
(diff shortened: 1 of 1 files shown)


www/docs/common-use-cases/semantic-search/custom_dimensions.md
```diff
@@ -95,10 +95,10 @@ END
 In order to calculate the final score of a document and query that contains
 custom dimensions, <Config v="names.product"/> takes the dot product of the
 query's custom dimensions with the document's custom dimensions and the resulting
-number is aded to their score.
+number is added to their score.
 
-Negative values decrease the overall score (sometimes caled "burying") and
-postive values increase the overal score (sometimes caled "boosting").  A dot
+Negative values decrease the overall score (sometimes called "burying") and
+postive values increase the overal score (sometimes called "boosting").  A dot
```
eskibars, line 101: `overal` here should be updated to `overall`

FYI, the reason I made these typos before was that I was editing through my tablet which has some quirky behavior when it has repeated keys (2 `l`s in a row, e.g.). There might be more of these types of typos

eskibars (changes requested):
One more change.  Thanks for the review/edit!

pwoznic (commented):
Done.

pwoznic (commented):
Updated other typos for overall, too, and the in this topic.

eskibars (approved):
LGTM, thanks!

PR #63, opened 2023-07-07
Update doc_meta example and curl command syntax

pwoznic:
Add doc_meta example and update curl command syntax with single quotes around URL


eskibars:
@pwoznic the build is failing.  It looks like this probably isn't a valid change from what I see.  Let's chat about it this week

pwoznic:
> @pwoznic the build is failing. It looks like this probably isn't a valid change from what I see. Let's chat about it this week

Sounds good. It doesn't like one of my examples.

eskibars (approved):
LGTM!  Thanks!

PR #64, opened 2023-07-09
Add core services to the APIs

eskibars:
There are lots of files changed here, but it's *relatively* straightforward change.  This adds the "core indexing services" to the OAS file and API playground


www/docs/rest-api/core-delete.api.mdx
amin2718, line 1: What's the difference between delete and core-delete? I thought we only had a single delete method.

www/docs/rest-api/create-corpus.api.mdx
amin2718, line 1: In create corpus, @nikron pointed out that some of the fields in the corpus object cannot or should not be set, i.e. they are ignored even if set. For example, dtProvision is one of these, but there might be others. I mention this because the sample Body includes these fields.

Furthermore, the filter attribute value sets an invalid type and level.

```
{
  "corpus": {
    "id": 0,
    "name": "string",
    "description": "string",
    "dtProvision": "string",
    "enabled": true,
    "swapQenc": true,
    "swapIenc": true,
    "textless": true,
    "encrypted": true,
    "encoderId": "string",
    "metadataMaxBytes": 0,
    "customDimensions": [
      {
        "name": "string",
        "description": "string",
        "servingDefault": 0,
        "indexingDefault": 0
      }
    ],
    "filterAttributes": [
      {
        "name": "string",
        "description": "string",
        "indexed": true,
        "type": "FILTER_ATTRIBUTE_TYPE__UNDEFINED",
        "level": "FILTER_ATTRIBUTE_LEVEL__UNDEFINED"
      }
    ]
  }
}
```

amin2718 (commented):
Looks good overall. I noticed a few issues.

www/docs/rest-api/core-delete.api.mdx
eskibars: I can delete this API, but this comes from the core_services proto file, which was listed as a separate endpoint.  I included it here for completeness, but I had the same question

tallatshafaat:
Thanks Shane. Looks fine. Answered your question, and added some minor comments.

tallatshafaat (approved):

www/static/vectara-oas.yaml
```diff
@@ -292,6 +294,62 @@ paths:
...
+        description: Request to delete a document from an index.
+        required: true
+      tags:
+        - IndexService
+      parameters:
+        - $ref: '#/components/parameters/customerIdentifier'
+      description: Core-Delete
+  /v1/core/index:
+    post:
+      operationId: Core-Index
```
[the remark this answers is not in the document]
tallatshafaat: Actually, `CoreIndex` might be better as it matches `FileUpload`.

www/docs/rest-api/create-corpus.api.mdx
eskibars: I agree, but this is is an effect of the upstream protos.  That is, https://github.com/vectara/protos/blob/main/admin.proto#L109 takes in a `Corpus` object, and a Corpus object has these attributes.  That same Corpus object is used in e.g. ListCorpora, where the return object correctly contains these parameters.

I think we have a few options here, but I don't think they should block this PR, since it's unrelated.  The options I see on the table are:
1. Split the Corpus object into request vs response objects in the upstream proto
2. A bit of Googling and subsequent reading led me to https://google.aip.dev/203 google.api.field_behavior, which looks like we could annotate these fields with  `[(google.api.field_behavior) = OUTPUT_ONLY]`
3. Just add some clarifying note in the proto description of which are settable and which are not.  We can custom craft examples, which is on the backlog of things to do anyway

@amin2718 @tallatshafaat what do you think

tallatshafaat: Yes, this is unrelated to this PR, so I don't think we should block this PR on it. I don't know why `create-corpus.api.mdx` was added to this PR. With the way the diff is shown in github, it is hard to see what changed in `create-corpus.api.mdx` (it is not part of core API, which this PR is about).

> Just add some clarifying note in the proto description of which are settable and which are not.
Some fields already have this, but somehow, I don't see it getting translated in the preview of this PR. For example, in our bitbucket Corpus proto, `dt_provision` has description:

```
  // The time at which the corpus was provisioned.
  // This value is ignored during Corpus creation.
```
But the preview for this PR doesn't show "This value is ignored during Corpus creation." line in the Playground.

image

www/static/vectara-oas.yaml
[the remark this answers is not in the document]
eskibars: Thanks, updated

eskibars:
I addressed all of the comments with the exception of https://github.com/vectara/vectara-docs/pull/64#discussion_r1257689128 which I think we can/should handle separately

www/docs/rest-api/create-corpus.api.mdx
eskibars: That's because the public protos haven't yet been merged with that update.  They were just pushed to dev, but now they need to go through all of the rest of the steps to get them here.  I've just opened https://github.com/vectara/protos/pull/16 to start that

PR #74, opened 2023-08-17
Update introduction and add new use case topic

pwoznic:
Updates the introduction and adds a new topic about use cases. It also updates the sidebar.


pwoznic:
Misc cleanup

www/docs/use-case-exploration.md
```diff
@@ -0,0 +1,124 @@
...
+This versatile <Config v="names.product"/> GenAI platform caters to a wide range of use 
+cases to drive better outcomes and unlock new possibilities in search 
+applications. <Config v="names.product"/> provides an easy entry point to generative AI 
+capabilities while protecting company IP and customer data. The data 
+is secure. <Config v="names.product"/> does not train on user data and respects data 
+sovereignty and provides you with peace of mind.
+
+:bulb: Check out [example Vectara applications, demos, and tutorials](http://vectara.com/demos/) to explore the many capabilities of our platform.
+
+### Conversational AI
```
cjcenizal, line 39: I had a hard time focusing on these as I read them. I think we can maintain interest by structuring the content so we first explain the value of the use case, and then follow it up with examples. I also think we can cut back on the hyperbole a bit, so the content becomes more meaningful. Here's how I might apply these ideas here:

```
Use <Config v="names.product"/> to provide interactive user experiences with
intelligent chatbots. These will enable your users to educate themselves
about your products and services through human-like interactions. The system
makes this possible by understanding the context of your users' questions and
using this context to create intelligent answers. The result is faster high-quality
answers for your users, and fewer escalations and support calls.

This technology takes the form of virtual assistants, chatbots, and messaging
applications for automating customer service and streamlining operations. 
```

What do you think of this approach and structure? I think if we applied it to the other sections below then folks will find the content more helpful.

www/docs/use-case-exploration.md
```diff
@@ -0,0 +1,124 @@
...
+
+We provide this list of sample <Config v="names.product"/> applications to help inspire developer 
+creativity:
+
+* [**AskNews**](https://asknews.demo.vectara.com)
+
+    Demonstrates the power of GenAI conversational search. Ask questions about news 
+    events around the world and get accurate answers thanks to <Config v="names.product"/>'s Grounded 
+    Generation.
+* [**Vectara Ingest**](https://github.com/vectara/vectara-ingest)
```
cjcenizal, line 107: Let's mention in the description that this is an open source project.

www/docs/use-case-exploration.md
```diff
@@ -0,0 +1,124 @@
...
+
+    Crawl datasets from pre-built crawlers (websites, RSS feeds, Jira, Notion, 
+    Docusaurus) and ingest them into Vectara
+* [**Vectara Answer**](https://github.com/vectara/vectara-answer)
+  
+    Enable users to create custom conversational search applications, such as
```
cjcenizal, line 113: Let's mention open source here too.

cjcenizal (approved):
LOVE IT! So excited to see these live!! I had some suggestions but nothing worth blocking on. Thank you for working on this Paul!

PR #78, opened 2023-08-23
Update H1/H2 headings throughout

pwoznic:
Fixed headings throughout the docs so that pages would not have multiple H1's


www/docs/common-use-cases/filtering-by-metadata/built_in_filterable_metadata.md
```diff
@@ -6,13 +6,13 @@ title: Default Filterable Metadata
...
 A few pieces of metadata are filterable out of the box, as they're very
-useful in a variety of situations.  This page describes these.
+useful in a variety of situations. This page describes these.
 
 Note that you can set up additional fields to filter on by setting up
 [filter attributes](/docs/api-reference/admin-apis/create-corpus#filter-attribute) on a
 corpus.
 
-## part.lang
+## part.lang field
```
cjcenizal, line 15: Maybe we should format this as code for consistency?

```
## `part.lang` field
```

www/docs/common-use-cases/filtering-by-metadata/built_in_filterable_metadata.md
```diff
@@ -24,7 +24,7 @@ Valid filter expressions for this would be something like:
 * `part.lang = 'deu'`
 * `part.lang = 'eng' OR part.lang = 'deu'`
 
-## part.is_title
+## part.is_title field
```
cjcenizal, line 27: Same here

cjcenizal (approved):
Looks great!! Had a couple non-blocking comments.

Also I noticed that the first page's "Not sure where to start?" heading appears as a child of "Language Agnostic", which seems like a mistake because the two aren't related. Should "Not sure where to start?" be an h2?

image

www/docs/common-use-cases/filtering-by-metadata/built_in_filterable_metadata.md
pwoznic: Thank you! I made the suggested updates in my latest commit

pwoznic:
Moved from draft to ready for review

PR #80, opened 2023-08-23
Update quickstart.md

pwoznic:
Made some minor changes, such as changing corpus->index.


www/docs/quickstart.md
```diff
@@ -126,18 +123,18 @@ lots of nuanced information about employee benefits.
...
 2. Enter a **Name** and **Description** such as _Employee Handbook_ and _The employee 
    handbook from HR_.
-3. (Optional) Add Filter Attributes to the corpus.
+3. (Optional) Add Filter Attributes to the index.
    
    Learn about [filter expressions](/docs/common-use-cases/filtering-by-metadata/filter-overview) if you want to use them in the future.
 4. Click **Create**. 
 
-   The **Corpus ID** displays next to the corpus title **Employee Handbook** in the 
+   The **Corpus ID** displays next to the index title **Employee Handbook** in the
```
cjcenizal, line 134: Should we replace "Corpus ID" with "Index ID"? If we do then we'll need to update the next sentence to be something like "If you use the API later you'll configure `corpus_id` fields with this value."

cjcenizal (approved):
Looks great! 1 nonblocking suggestion.

PR #83, opened 2023-08-28
Update key concepts

pwoznic:
Added this new topic in the Getting Started to provide an overview of conceptual topics in Gen AI. As we build out the new concepts chapter (Learn), we will link to the appropriate overview section for each term. I also created supporting diagrams for each section. I would like to create a more complex diagram that ties everything together in a future PR.


www/docs/key-concepts.md
```diff
@@ -0,0 +1,179 @@
...
+### Summarization 
+
+Summarization algorithms utilize Natural Language Processing (NLP) techniques 
+to distill large volumes of data into concise summaries. This reduced 
+information enables rapid data assimilation and decision-making
+By leveraging machine learning models, summarizations adapt to different 
+contexts and content types, while preserving the essence of the original 
+data
+
+### Hallucinations
```
cjcenizal, line 97: How about changing this to "Reducing hallucination", since "Hallucination" isn't exactly a technique?

www/docs/key-concepts.md
```diff
@@ -0,0 +1,179 @@
...
+
+Hallucinations occur when AI generates incorrect information that was not 
+present in the input or training data. Simply put, chatbots tend to make
```
cjcenizal, line 100: I suggest removing the "Simply put..." sentence because I think it's an oversimplification and potentially misleading. The rest of the paragraph sounds accurate to me.

cjcenizal (changes requested):
I think we should make some changes to this before merging. @amin2718 Would you mind reviewing my comments to ensure my takes are accurate?

I think we also need to adjust the AI Architecture image, to make the relationships between the boxes clearer. To move quickly we can just remove the lines connecting the boxes, but we can refine it further by making it clear that "Vector Databases" enable the various types of search, and then feed the result to "LLMs" to provide helpful answers to questions.

pwoznic:
> I think we should make some changes to this before merging. @amin2718 Would you mind reviewing my comments to ensure my takes are accurate?
> 
> I think we also need to adjust the AI Architecture image, to make the relationships between the boxes clearer. To move quickly we can just remove the lines connecting the boxes, but we can refine it further by making it clear that "Vector Databases" enable the various types of search, and then feed the result to "LLMs" to provide helpful answers to questions.

Sounds good. I originally wanted to create a diagram that covered everything but I thought it would be helpful to create a diagram for each section and we can discuss as a team on the best way to present the info, combine, etc. I will go through comments on what I can change.

pwoznic (commented):
I made changes based on the comments

pwoznic:
We're working on a new key concepts now so closing this one.

PR #84, opened 2023-09-01
Add API Quickstart topic

pwoznic:
Added a quick start topic for the API called API Jumpstart. The intent is to show a variety of basic queries around searching, uploading files, updating metadata, listing indices, etc. This topic does not include humorous examples like the console quickstart because I started this draft a while ago. I can make the content more engaging and entertaining if necessary.

My vision is to have additional examples throughout the documentation that are more advanced and showcase more real-world situations that explain what we're doing and why, e.g. more examples about metadata.

### DIFF
--- a/www/docs/api-jumpstart.md
+++ b/www/docs/api-jumpstart.md
@@ -244,8 +244,109 @@ curl -L -X POST 'https://api.vectara.io/v1/index' \
 
 
 
-## Index the corpus
+## Query a term and return a specific number of results
 
+In this example, you want to search for the term "technology," and then return 
+only the first 5 results. 
+
+### Example cURL command
+
+```json
+
+curl -X POST "https://api.vectara.io/v1/query" \
+     -H "x-api-key: 12345" \
+     -H "customer-id: 7890" \
+     -d '{
+         "corpusId": "2",
+         "query": "technology",
+         "numResults": 5,
+         "start": 0
+     }'
+
+
+
+```
+
+### Example response with 5 results
+
+```json
+
+{
+  "status": "OK",
+  "results": [
+    {
+      "text": "The future of technology is AI.",
+      "score": 0.98,
+      "documentIndex": 1
+    },
+    {
+      "text": "Technology is evolving rapidly.",
+      "score": 0.95,
+      "documentIndex": 2
+    },
+    {
+      "text": "Blockchain technology is revolutionary.",
+      "score": 0.92,
+      "documentIndex": 3
+    },
+    {
+      "text": "Technology has its pros and cons.",
+      "score": 0.90,
+      "documentIndex": 4
+    },
+    {
+      "text": "The role of technology in modern society.",
+      "score": 0.88,
+      "documentIndex": 5
+    }
+  ]
+}
+
+```
+
+## Query with metadata filters
+
+In this example, you want to query _cloud computing_ with 10 results in 
+the _IT_ category.
+
+### Example cURL Command
+
+```json
+curl -X POST "https://api.vectara.io/v1/query" \
+     -H "x-api-key: 12345" \
+     -H "customer-id: 7890" \
+     -d '{
+         "corpusId": "2",
+         "query": "cloud computing",
+         "numResults": 10,
+         "metadataJson": {
+             "category": "IT"
+         }
+     }'
+
+
+### Example query response
+
+
+```json
+
+{
+  "status": "OK",
+  "results": [
+    {
+      "text": "Cloud computing is changing the IT landscape.",
+      "score": 0.99,
+      "documentIndex": 8
+    },
+    {
+      "text": "The benefits of cloud computing are numerous.",
+      "score": 0.96,
+      "documentIndex": 9
+    }
+  ]
+}
+
+```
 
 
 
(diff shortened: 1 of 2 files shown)


www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
+---
+id: api-jumpstart
+title: API Jumpstart
```
cjcenizal, line 3: Can I suggest calling these "API recipes" or "App recipes" (I prefer this one) to avoid confusion between Quickstart and Jumpstart? I think the idea of "recipes" describes these nicely... you follow a few easy steps and you achieve a delicious outcome.

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+sidebar_label: API Jumpstart
+---
+import {Config} from '@site/docs/definitions.md';
+
+Using the API enables you to integrate the <Config v="names.product"/> search capabilities into your
```
cjcenizal, line 8: We publish numerous APIs, not just a single one. So for consistency, I suggest referring to "our APIs" or "Vectara's APIs" instead of "the API".

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
+---
+id: api-jumpstart
+title: API Jumpstart
+sidebar_label: API Jumpstart
+---
+import {Config} from '@site/docs/definitions.md';
+
+Using the API enables you to integrate the <Config v="names.product"/> search capabilities into your 
+applications. While the console UI offers a user-friendly interface for initial 
+testing and manual operations, the <Config v="names.product"/> API is where the
```
cjcenizal, line 10: The Console UI and our APIs work hand-in-hand as part of the app development process. For example, a developer will:

1. Fine-tune a query's lambda and filters until the answer quality is just right.
2. Click "copy request" button and paste it into their IDE.
3. Copy their customer ID and API key from Console to further configure the request.
4. Test out their software and then verify requests are hitting their corpus by checking the querier graph on the Overview tab.

Generally, I think we should be careful to avoid casting Console as "less than" the APIs, because they are both part of a single product.

For this particular sentence, maybe we can just remove it?

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+* `x-api-key` = abc_12345defg67890hij09876
+* `corpus_id` = 1
+* `query` = How much PTO is offered to employees each year?
+
+
+### Example cURL command
+
+This example queries the index with the question about annual PTO.
+
+```json
```
cjcenizal, line 57: Oddly enough this highlights pretty nicely with `js`.

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+    }
+  ]
+}'
+```
+
+### Example JSON response
+
+Let’s take a closer look at the first response:
+
+```json
```
cjcenizal, line 113: I think this will highlight more nicely as `js`.

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+    }
+  ]
+}
+```
+## Create a new index and ingest a document
+
+In this example, you want to create a new index and also ingest a document in 
+this new index.
+
+1. Execute the following cURL command:
```
cjcenizal, line 436: In previous examples this is an h3. Should format this the same for consistency?

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+   ```json
+   curl -X POST "https://api.vectara.io/v1/create-corpus" \
+     -H "x-api-key: abc_12345defg67890hij09876" \
+     -H "customer-id: 123456789" \
+     -d '{
+         "name": "AI Future"
+     }'
+   ```
+
+ You get the following response:
```
cjcenizal, line 447: Same here.

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+}
+```
+## Create a new index and ingest a document
+
+In this example, you want to create a new index and also ingest a document in 
+this new index.
+
+1. Execute the following cURL command:
+
+   ```json
```
cjcenizal, line 438: As with the other examples, this seems to format more nicely with `js`.

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+   {
+  "status": "OK"
+   }
+   ```
+
+
+## List all indices and delete a specific index
+
+1. Execute the following curl command to list the indices:
```
cjcenizal, line 480: Same comment here and elsewhere about using an h3.

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+   ```
+  Notice that you only have 2 indices now.
+
+## Update a document, reindex the document, and query the index
+
+In this example, you want to update a document with new information. Reindexing 
+the document improves data retrieval time when you ask a query. After the 
+reindex, you want to query the index and get a result from the updated information.
+
+1. Execute the following curl command to update an existing document in your
```
cjcenizal, line 540: Not sure what's going on here but it looks like GitHub is getting confused about how to parse the Markdown here.

image

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,672 @@
...
+  ]
+}
+   ```
+
+In this final example, you updated an existing document to include new 
+information from the latest version of a document. You wanted to reindex 
+this updated document to ensure that the latest content is searchable. 
+Finally, you issued a query that asked a question about the new content.
+
+Thie API jumpstart provided a variety of query examples that you can leverage
```
cjcenizal, line 671: Typo: "Thie" -> "This"

And if we go with "recipes" then we'll need to update the reference to "jumpstart".

cjcenizal (changes requested):
This is fantastic content! I spotted some formatting problems I think we need to dig into and also had some other suggestions about content.

pwoznic (commented):
Addressed changes from comments

www/docs/api-jumpstart.md
```diff
@@ -0,0 +1,683 @@
...
+
+     You get the following response:
+
+     ```js
+     {
+  "status": "OK"
+     }
+   ```
+
+3. Execute the curl command from Step 1 again:
```
cjcenizal, line 530: Since there's no code snippet after this sentence, maybe it should be inlined with the following sentence?

```
3. Execute the curl command from Step 1 again. You get the following response:
```

cjcenizal (approved):
Looks good! I still see some weird formatting issues in the preview, but maybe they won't show up in Docusaurus.

pwoznic (commented):
Updated step 3

PR #85, opened 2023-09-06
Fix app header z index

3eif:
Docusaurus sets the app header's z index to 200 by default. This PR sets it back to it's proper value, 8.


cjcenizal (approved):
Thanks for fixing this! Tested locally and it doesn't look like this breaks anything in the layout.

@3eif While you're here could you add a rule to `[role="banner"]` on line 188?

```css
[role="banner"] {
  z-index: 1;
  /* snip */
}
```

This will fix a small bug in the UI:

image

3eif:
> @3eif While you're here could you add a rule to [role="banner"] on line 188?

Done!

3eif:
Can you merge this when possible? I don't have permission on this repo

PR #102, opened 2023-10-05
Add MMR reranker content

pwoznic:
Updated Reranking topic with info about MMR


www/docs/api-reference/search-apis/reranking.md
```diff
@@ -1,25 +1,47 @@
...
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 import {Config} from '@site/docs/definitions.md';
 import {vars} from '@site/static/variables.json';
 
-Reranking is the process by which an initial set of query results is rescored
-to produce a more refined and accurate ranking. It's usually performed with a
+Reranking is a process by which an initial set of query results is rescored 
+to produce a more refined and accurate reanking. It's usually performed with a
```
amin2718, line 13: Reranking is misspelled. This is a nit, but it's a bit redundant to say results are rescored to produce a more refined reranking. You can just say "rescored to produce a more refined ranking".

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -1,25 +1,47 @@
...
 
 Reranking is a beta feature in the platform, and only available as a single
 reranker for English results, and can be enabled by specifying 272725717 as
 the `reranker_id`.
 
-In order to use the reranker you need to specify two additional parameters in
-the request: `start` and `numResults`.  These tell <Config v="names.product"/>
-which results to rerank.  In most scenarios, it makes sense to set `start` to
+To use the reranker, you specify two additional parameters in
+the request: `start` and `numResults`. These parameters tell <Config v="names.product"/>
```
amin2718, line 22: This sentence isn't really accurate. Every query request must specify these, although if they're not specified, start defaults to 0 and numResults to 10, I think. Or maybe numResults is required, I can't remember.

The key point is that this has nothing to do in particular with reranking, which is what the sentence seems to be implying. On the other hand, the next sentence is quite accurate. Setting start to anything other than 0 is probably not the right thing to do when using a reranker.

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -1,25 +1,47 @@
...
-`numResults` to the total number of documents you wish to rerank.
\ No newline at end of file
+`numResults` to the total number of documents you wish to rerank.
+
+## Maximal Marginal Relevance Reranking
+
+The Maximal Marginal Relevance (MMR) Reranker enables you to diversify search 
+results to reduce redundancy while maintaining relevance to the query. Having 
+a diverse set of relevant results can help ensure that users get answers from 
+a many sources of data.
```
amin2718, line 32: Extra "a" at the beginning of the line.

I think the statement of benefit is slightly off. I would say  having a diverse set of relevant results has different benefits depending on the use case:

1. In a pure search scenario, it improves user engagement with results by avoiding repetition.
2. In a generative AI scenario, it produces more comprehensive summaries
3. Closely related to these points, it can be seen as potentially representing all points of view in the data or reducing bias.

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -1,25 +1,47 @@
...
+In addition to specifying the `reranker_id` at query time, you also 
+specify a `diversity bias` range between `0` and `1`. You can also enable the 
+MMR Reranker in the console UI as follows:
+
+1. Open a corpus from the list and select the **Search** tab.
+2. Click **Advanced Options** and a navigation drawer opens.
+
+   ![Diversity Reranker](/img/diversity_reranker.png)
+3. Enter a value between `0` and `1` in the `Diversity factor` field. Values 
+   closer to 1 are optimized for the most diverse results.
```
amin2718, line 43: When values are floating point, I prefer to see them represented as 0.0 and 1.0, not 0 and 1. I think it just makes it a little clearer to the users.

I would also say "Values closer to 1 optimize for the most diverse results" which is more direct than "are optimized for".

amin2718 (changes requested):
Looks good. Left some suggestions.

www/docs/api-reference/search-apis/reranking.md
pwoznic: Done

www/docs/api-reference/search-apis/reranking.md
pwoznic: Rewrote it to be more accurate

www/docs/api-reference/search-apis/reranking.md
pwoznic: Added these points

www/docs/api-reference/search-apis/reranking.md
pwoznic: Done

pwoznic (commented):
Thanks, Amin! I made changes based on your feedback.

amin2718 (approved):
Thanks Paul

PR #111, opened 2023-10-13
Update client app screens

pwoznic:


pwoznic:
PR 110 reverted the redirects file and 3 screen captures by mistake so reapplying them here.

PR #117, opened 2023-10-19
Update topic with maxSummarizedResult = 10

pwoznic:
Updated the topic and gave a brief explanation about not maximizing the value. Let me know if I should elaborate more on using different values (e.g lower, mid, and high/max value).


ofermend (approved):
LGTM

eskibars (approved):

PR #122, opened 2023-10-25
[feat] add Search component

mrderyk:
**CHANGES**
- implement Search component
  - implement SearchInput
  - implement SearchResults
- introduce temporary internal test page

**KNOWN ISSUES**
- modal closes when click border around search input (might be better to implement "click out to close" in `VuiModal`

**SCREENSHOTS**
https://github.com/vectara/vectara-docs/assets/1464245/d6f46fe1-32a3-47a7-9e22-f76601ceb581


vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const { fetchSearchResults, isLoading } = useSearch(
+    customerId,
+    corpusId,
+    apiKey,
+    apiUrl
+  );
+
+  const sendSearchQuery = async (query: string) => {
+    if (isLoading) {
+      return;
```
cjcenizal, line 55: This is a corner case, but what if the user adds some more input and hits Enter to send a query while this is loading? Won't the results be stale? I think the same thing will happen if the user enters input while this is in a loading state (even without hitting Enter).

I just tested this by:

1. Setting my Network throttling to 3G (Fast)
2. Copying the string " is vectara" into my clipboard
3. Entering the input "what"
4. Waiting until the spinner was visible and then quickly pasting " is vectara" to complete the input

The search doesn't re-execute for the completed query, so the search results appear stale.

I think we can solve this by removing this condition, tagging each request with an incrementing ID, and ensuring only the latest request results in an update to the UI. I tested this locally and it seemed to work but I'd like to know what you think.

```ts
// ln 21
let searchCount = 0;

/* snip */

const sendSearchQuery = async (query: string) => {
  const searchId = ++searchCount;

  const results = await fetchSearchResults(query);

  // If we send multiple requests in rapid succession, we only want to
  // display the results of the most recent request.
  if (searchId === searchCount) {
    setSearchResults(results);
    setSelectedResultIndex(null);
    selectedResultRef.current = null;
  }
};
```

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+      if (key === "Escape") {
+        closeModalAndResetResults();
+      }
+
+      if (key === "Enter") {
+        evt.preventDefault();
+
+        if (selectedResultIndex !== null) {
+          window.open(searchResults[selectedResultIndex].url, "_self");
```
cjcenizal, line 88: Do we need this? When I use the tab keys to navigate the modal, it seems like the default browser accessibility features enable me to use Enter to open the links.

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,206 @@
...
+        } else {
+          sendSearchQuery(queryRef.current);
+        }
+      }
+
+      if (searchResults.length === 0) {
```
cjcenizal, line 94: For the same reason as above I think we can remove lines 94-112.

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,206 @@
...
+    queryRef.current = evt.target.value;
+    debouncedSendSearchQuery(evt.target.value);
+  };
+
+  const onKeyDown = useCallback(
+    (evt: KeyboardEvent) => {
+      const key = evt.key;
+
+      if (key === "Escape") {
+        closeModalAndResetResults();
```
cjcenizal, line 81: Looks like this doesn't work if focus is on a search result. Can we add the handler for this event to the document itself when the modal is opened, and remove the handler when the modal is closed? Maybe the right place to do this is inside `SearchModal.tsx`.

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,206 @@
...
+            >
+              <SearchResult
+                searchResult={searchResult}
+                isSelected={selectedResultIndex === index}
+              />
+            </div>
+          );
+        });
+
+  const onBodyClick = useCallback((evt: MouseEvent) => {
```
cjcenizal, line 147: We can simplify things by implementing a change in `SearchModal.tsx` on line 49:

```tsx
<VuiScreenBlock onClick={() => onClose()}>
```

We should also update `onClose` to be a required prop in that component.

Then we can implement `onClose` in this file:

```tsx
<SearchModal
  ref={modalRef}
  isLoading={isLoading}
  onChange={onChange}
  onKeyDown={onKeyDown}
  isOpen={isOpen}
  onClose={() => closeModalAndResetResults()}
  resultsList={resultsList}
/>
```

This will make the `onBodyClick` function unnecessary.

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+      if (shouldCloseModal) {
+        closeModalAndResetResults();
+      }
+    }
+  }, []);
+
+  useEffect(() => {
+    if (selectedResultRef.current) {
+      selectedResultRef.current.scrollIntoView({
```
cjcenizal, line 165: Per my comment about default browser accessibility, I think this comes for free so we can remove this useEffect.

cjcenizal (approved):
Had some suggestions on ways to simplify the code, but nothing critical. I found one corner-case bug. Overall, LGTM!

vectara-docusaurus-search/src/components/search/Search.tsx
mrderyk: Yes I think it'd be nice to keep it in as an alternate way of navigating through the results. For a site like Google, for instance, the way to navigate results is by up/down arrows as opposed to tabs, so supporting this paradigm would be good for usability, IMO.

vectara-docusaurus-search/src/components/search/Search.tsx
mrderyk: Yeah I think I saw this somewhere in the Vectara Answer codebase. I'll check it out!

vectara-docusaurus-search/src/components/search/Search.tsx
cjcenizal: OK got it. I might have broken it by accident because I'm not able to use the keys to navigate the items -- sorry!

vectara-docusaurus-search/src/components/search/Search.tsx
mrderyk: https://github.com/vectara/vectara-docs/pull/122/commits/c0d55f422d810a924da848d57e739520dcb53180

vectara-docusaurus-search/src/components/search/Search.tsx
mrderyk: https://github.com/vectara/vectara-docs/pull/122/commits/c0d55f422d810a924da848d57e739520dcb53180

I ended up just removing this, since `onEscapeKey={onCloseDelayed}` in `SearchModal` already seems to do the trick!

vectara-docusaurus-search/src/components/search/Search.tsx
mrderyk: https://github.com/vectara/vectara-docs/commit/c0d55f422d810a924da848d57e739520dcb53180

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const queryRef = useRef<string>("");
+  const searchCount = useRef<number>(0);
+  const { fetchSearchResults, isLoading } = useSearch(
+    customerId,
+    corpusId,
+    apiKey,
+    apiUrl
+  );
+
+  const sendSearchQuery = async (query: string) => {
```
cjcenizal, line 52: I think we should trim `query` and check if it's an empty string. If it is empty I think we should skip sending the request and just render an empty state. Currently, if you enter a query and then delete it you end up in this state:

image

mrderyk: Hmm yeah I agree.

At first, I thought to allow the empty string so the user would at least see "something" there instead of going back to just the search input in the very center of the screen. This was due to us not locking the search to the top, so it was pretty jarring once the search results got removed.

But thanks to your new change, we can do this right. :)

PR #128, opened 2023-10-26
[feat] Combine multiple search results by URL.

mrderyk:
**CHANGES**
- Combine multiple results the belong to the same URL
- Re-order combined results by averaging the relevance scores of documents belonging to each URL.

### DIFF
--- a/vectara-docusaurus-search/src/components/search/useSearch.tsx
+++ b/vectara-docusaurus-search/src/components/search/useSearch.tsx
@@ -32,7 +32,7 @@ export const useSearch = (
           {
             query,
             start: 0,
-            numResults: 10,
+            numResults: 20,
             corpusKey: [
               {
                 corpusId,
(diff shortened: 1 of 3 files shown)


vectara-docusaurus-search/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2016",
+    "target": "es2017",
```
mrderyk, line 3: Had to do this so TS would properly recognize `Object.entries`.

According to [this reference of features introduced in ES2017](https://caniuse.com/async-functions,object-values,object-entries,mdn-javascript_builtins_object_getownpropertydescriptors,pad-start-end,mdn-javascript_grammar_trailing_commas_trailing_commas_in_functions), the only browser potentially affected is IE11, which is EOL anyway.

vectara-docusaurus-search/src/components/search/useSearch.tsx
```diff
@@ -125,3 +130,45 @@ export const parseSnippet = (source: string) => {
       : [textAndPost, ""];
   return { pre, post, text };
 };
+
+const compileWeightedResults = (
```
cjcenizal, line 134: Hmmm I'm not sure I understand the intended UX behind this. Why is the average of the scores for duplicate results the correct score for comparison across deduplicated results? Why not, say, the sum?

Or even more simply, why not ignore scores at this point? If we add up the number of duplicate result and then sort the deduplicated results based on these sums, does that yield a desirable UX?

cjcenizal (approved):
So simple! Tested locally and it works great. Code LGTM.

PR #132, opened 2023-10-27
feat: Implement search plugin in local folder

mrderyk:
**CHANGES**
- implement plugin widget JS code that imports the `Search` component
- implement Docusaurus plugin to load widget into the docs site
- implement CTRL + K to open search bar
- clean up SCSS imports to avoid style regressions


vectara-docusaurus-search/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "es2020",
```
mrderyk, line 3: This is to allow optional chaining syntax.

vectara-docusaurus-search/src/components/search/_index.scss
```diff
@@ -1,4 +1,17 @@
-@import "../../../vui/index";
+@import "../../../vui/components/index";
```
mrderyk, line 1: The reset breaks the styling for elements not related to search.
This is what was causing spacing, font size, etc issues on the page.

Removing the overall reset fixes the issue and retains styles we put into search-related components.

We should think about how this might affect users that import our components in the future though. 

i.e. If we make `<Search>` available to customers in the future and it requires a reset, how do we avoid affecting a user's existing styles? Must we always wrap exported components in a web component? How do other component libraries do it (e.g. Chakra UI)? 

Components I've exported via component libraries before used things like Emotion, Styled Components, or were built off existing libs like Chakra, so I'm not sure yet on how our exports would work.

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -81,8 +82,14 @@ export const Search: FC<Props> = ({ customerId, apiKey, corpusId, apiUrl }) => {
     debouncedSendSearchQuery(currentQuery);
   };
 
+  const openSearchOnKeyStroke = (e: KeyboardEvent) => {
```
cjcenizal, line 85: Nit: I like to localize code as much as possible. Since this is only used inside of the `useEffect` on line 171, how about placing it inside? This way it's clear to the reader that it's only consumed there.

vectara-docusaurus-search/src/components/search/_index.scss
cjcenizal: > Must we always wrap exported components in a web component?

This seems like the obvious solution to me. Styles-in-JS are also an option, but that's a much bigger change so I'd be curious about how the tradeoffs compare. Is such a big change worth it?

vectara-docusaurus-search/src/plugin/index.tsx
```diff
@@ -0,0 +1,46 @@
+import { createRoot } from "react-dom/client";
+import { Search } from "../components/search/Search";
+
+(function appendWidget() {
+  // When the search container has been added to the DOM,
+  // add the search component to it.
```
cjcenizal, line 6: Thanks for adding all of these comments!

vectara-docusaurus-search/src/plugin/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  var observer = new MutationObserver(() => {
```
cjcenizal, line 7: So cool! I haven't seen MutationObservers used much.

www/package.json
```diff
@@ -18,8 +18,7 @@
     "@docusaurus/core": "2.2.0",
     "@docusaurus/preset-classic": "2.2.0",
     "@mdx-js/react": "^1.6.22",
-    "@vectara/docusaurus-search": "file:../docusaurus-search",
-    "@vectara/plugin-vectara-search": "file:../vectara-docusaurus-search",
+    "@vectara/docusaurus-search": "file:../vectara-docusaurus-search",
```
cjcenizal, line 21: Neat! TIL.

cjcenizal (approved):
Tested locally, code LGTM!

vectara-docusaurus-search/src/components/search/_index.scss
```diff
@@ -15,3 +13,9 @@ button {
   border-radius: 0;
   font-size: inherit;
 }
+
+.searchModalContainer {
+  @import "../../../vui/_reset";
+  @import "./searchInput";
+  @import "./searchResult";
```
mrderyk, line 20: Basically: Only have the reset take effect inside the search modal- anything else on the docs site stays the same.

cjcenizal: 👍 This won't change any styles from the docs site that are affecting the search component, though, right?

pwoznic:
Can the styling of the search button be modified to stand out more? For example, the search from Algolia "pops" more on the page with a wider box, and it also shows the keyboard shortcut:

image

More captures of their flow (wider pop-up):

image

Search results styling (shows navigation and also renders a preview):

cjcenizal:
> Can the styling of the search button be modified to stand out more? 

This is a great idea! I can implement this. I also super-appreciate your other research @pwoznic. 🙏 I extracted your last point to https://github.com/vectara/vectara-docs/issues/142 so we can address it later.

PR #143, opened 2023-11-01
Replace Grounded Generation with Retrieval Augmented Generation

pwoznic:
Updated topics to refer to RAG instead of GG


eskibars (approved):
LGTM, thanks

PR #146, opened 2023-11-02
Added Snow Analytics code.

aamirbutt:
This add Snow Analytics code to the docs website. 
It appears that the rudderstack analytics code that we have previously might be incomplete, and is sending the events only on first time load. We can fix that, but we will get rid of that eventually, and move all the analytics to snowAnalytics instead.

### DIFF
--- a/www/static/snowAnalytics.js
+++ b/www/static/snowAnalytics.js
@@ -79,7 +79,8 @@
   loadLib();
 
   whenAvailable("_analytics", function(_analytics) {
-    var Analytics = _analytics.init({
+    // Storing in window so we can access it from the routeUpdateModule
+    var Analytics = window.snowAnalytics = _analytics.init({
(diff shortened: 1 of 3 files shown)


www/static/snowAnalytics.js
```diff
@@ -0,0 +1,112 @@
...
+              },
+              body: pageJSON
+            });
+          }
+        }      
+      ]
+    });
+    Analytics.page();
+    // Storing in window so we can access it from the routeUpdateModule
+    window.snowAnalytics = Analytics;
```
cjcenizal, line 110: Can we move this to line 82, to colocate assignment?

```js
var Analytics = window.snowAnalytics = _analytics.init({
```

cjcenizal (approved):
Tested locally, code LGTM!

PR #147, opened 2023-11-02
Add TopicButton, Grid, and Spacer components.

cjcenizal:
@pwoznic I added some components and used them to drive people to the Quick Start and API Playground from the "Get Started" page. I also changed the heading of this page to "The Vectara Platform" and made the initial introduction phrasing more direct because it seemed to flow a bit better. This is just an idea of what's possible with these components, so if you want to try something different with them just LMK!

image


cjcenizal:
CC @eskibars because you expressed interest.

pwoznic (approved):
Looks great!

PR #163, opened 2023-11-08
fix: ensure rendering a single instance of search

mrderyk:
**PROBLEM**
The original usage of MutationObserver was intentionally implemented without disconnecting, in order to allow the search UI to be mounted on route change. **This is because the nav bar gets re-rendered on navigating to a new docs page.** However, this was causing multiple instances of the Search component to be mounted on the page. Clicking the button would only activate one, using the hot key would cause all instances to open.

***Note: I'm not quite sure about the inner workings of this, as the search container only contains one button, which would initially point to just a single instance of the Search being mounted. But using the hot key proves the existence of multiple search components, as they all respond to the hot key and open together. I suspect this is something related to creating a root out of the search container and mounting to it multiple times.***

**CHANGES**
- Disconnect mutation observer on mount of search (ensures a single instance)
- Use Docusaurus "onRouteUpdate" lifecycle method to re-init mutation observer (ensures re-mount of search on new navbar instance)

**SCREENSHOTS**

BEFORE
modal_before

AFTER
modal_after


vectara-docusaurus-search/plugin.js
```diff
@@ -1,5 +1,20 @@
...
 }
+
+export default (function () {
+  if (!ExecutionEnvironment.canUseDOM) {
+    return null;
+  }
+  const event = new Event("onRouteUpdated");
+  return {
+    onRouteDidUpdate() {
+      document.dispatchEvent(event);
```
cjcenizal, line 17: Nit: This looks like it dispatches a reference to the same event instance. Is that deliberate?

I have no idea if that will be problematic, but it seems more "correct" to instantiate a new event here, at the moment of dispatch. Conceptually, this method is intended to indicate a new event has occurred, so it seems truer to this concept to create a new event.

```js
document.dispatchEvent(new Event("onRouteUpdated"));
```

vectara-docusaurus-search/src/plugin/index.tsx
```diff
@@ -29,17 +31,37 @@ import { Search } from "../components/search/Search";
...
   });
 
-  observer.observe(document, {
-    attributes: false,
-    childList: true,
-    characterData: false,
-    subtree: true,
-  });
+  /**
+   * Whenever docusaurus changes route, the nav bar gets rerendered.
```
cjcenizal, line 48: Oh, wow! Was this the cause of the original bug? Really amazing. Seems tricky to debug. FWIW I would love to see this kind of summary in the PR description. This is typically the context I'm looking for when I do a code review or looking at old PRs (for forensic reasons, for example).

cjcenizal (approved):
Tested locally, code LGTM. Great work figuring this one out!

vectara-docusaurus-search/src/plugin/index.tsx
mrderyk: Oh no this was an adjacent issue. But I did update the PR comment to give more context.

vectara-docusaurus-search/plugin.js
mrderyk: The event properties never really change, hence I just decided to use the same event. You make a good point though. I'll change it.

mrderyk: [21a88c1](https://github.com/vectara/vectara-docs/pull/163/commits/21a88c14e8915471c662ec6bfae6b612e1ed374d)

PR #167, opened 2023-11-13
feat: store/retrieve search results from local storage

mrderyk:
### CONTEXT
We need to store search history for our search component, and ensure that recorded entries are unique for every search box. As a first step, we create a hook to write/read previous search terms to/from local storage. These will be passed to the search modal in a follow-up step.

### CHANGES
- Create a search history hook that exposes read/write functions to the consumer.
- On search, store the latest term and discard previous terms that are outside the history limit (default is 10).

### SCREENSHOTS
- None, data changes only for now


mrderyk:
@cjcenizal Just addressed the conflict. Good to review now!

vectara-docusaurus-search/src/components/search/Search.tsx
```diff
@@ -25,22 +26,53 @@ import { SearchResult } from "./SearchResult";
...
+  apiUrl,
+  historySize = 10,
+}) => {
   const [searchResults, setSearchResults] = useState<
     DeserializedSearchResult[]
   >([]);
 
+  // Compute a unique ID for this search component.
+  // This creates a namespace, and ensures that stored search results
+  // for one search box don't appear for another.
```
cjcenizal, line 65: Can you walk me through the user experience problem this is solving? I'm not following the bit about multiple search boxes. Do you mean if there are multiple tabs open on the docs site, each of which has its own search box?

cjcenizal (commented):
Looks good overall! But I have a question about the problem we're trying to solve. Can we talk through that bit before merging?

vectara-docusaurus-search/src/components/search/Search.tsx
mrderyk: Steps to test:
1. Run app locally.
2. Do some searches.
3. Open the console, go to "Application", on the left hand side open "Local Storage". Under "localhost:5000", find the most recent search history:
Screenshot 2023-11-20 at 11 44 26 AM

4. Under `www/docusaurus.config.js`, change the search configuration to point to a different customer ID, corpus ID, and api key (this doesn't have to be a legit config- just something else to point the search box to)
Screenshot 2023-11-20 at 11 48 15 AM

5.  Rebuild the app and perform searches against the updated search. Then view history under console:
Screenshot 2023-11-20 at 11 51 28 AM

cjcenizal (approved):

PR #168, opened 2023-11-13
added console button

ofermend:
Added a button to the console in the navbar


cjcenizal (approved):
This is great! Can we position the link on the right? I think it's more symmetrical with the API docs links in Console.

image

image

ofermend:
Yes, done!

PR #171, opened 2023-11-17
chore: update readme

mrderyk:
### CONTEXT
The readme doesn't specify the correct directory from which to run the commands to:
- run the docs site locally
- build the docs site
- serve the docs site from the development machine

The readme needs to be updated to reflect this.

### CHANGES
- Update readme to reflect the correct directory from which to execute commands
- Catch analytics CORS fetch errors so that the site running in dev mode doesn't show the error overlay
  - ideally, we should only run the analytics fetch if NODE_ENV is not `development`, but that env var doesn't seem available from the context of additional scripts set up in the docusaurus config.


cjcenizal (approved):
LGTM!

eskibars (approved):

PR #176, opened 2023-11-24
Add update filter attributes and list jobs

eskibars:
The main goal of this PR was to add the new list jobs and update filter attributes APIs to the OpenAPI spec (and thus the API playground).  However, I've also used it to give each API a "real" description, including a very brief set of tips for using the API based on stumbling blocks we've seen users run into.

All of the changes for this PR lie under the API playground, so for review please check there (it's hard to compare the source auto-generated .mdx files, I recommend you use the deploy preview vs current production and just comment on this issue if anything feels off).

We'll also want more long-form content for these APIs, but I wanted to make sure we had at least API spec docs in place before we launch.


www/static/vectara-oas.yaml
```diff
@@ -52,7 +52,16 @@ paths:
...
       parameters:
         - $ref: '#/components/parameters/customerIdentifier'
-      description: Create Corpus
+      description: |
+        Creates a corpus, which is a container to store data in.
+
+        Some tips for this API:
+        * This operation only works with OAuth 2.0 (in a JWT "Bearer Token") authentication. It does not work with API Keys.  You can find details of how to set up and use OAuth 2.0 [here](https://docs.vectara.com/docs/learn/authentication/oauth-2).
+        * The `name` of the corpus is the only required field.
+        * Filter attributes tell Vectara which metadata fields you'd like to run SQL-style filters against.  If you need to change them after you've created the corpus, see the [Replace Filter Attributes API](https://docs.vectara.com/docs/rest-api/replace-corpus-filter-attrs)
```
tallatshafaat, line 61: nit: Curious, are the hyper linked supposed to the relative path? When viewing the deploy preview, the links weren't to the deploy preview. Not important though.

www/static/vectara-oas.yaml
```diff
@@ -176,7 +197,14 @@ paths:
         - IndexService
       parameters:
         - $ref: '#/components/parameters/customerIdentifier'
-      description: Index
+      description: |
+        This is the "standard" Indexing API for indexing semi-structured, text-heavy "documents."  Indexing data into Vectara is typically very fast: within a few seconds.
```
tallatshafaat, line 201: nit: move full stop outside quotes?

www/static/vectara-oas.yaml
```diff
@@ -224,15 +293,64 @@ paths:
...
+        * By default, Vectara only uses its neural/semantic retrieval model, and does not attempt to use keyword matching.  To enable hybrid search with a mix of both keyword and neural results, edit the `lambda` value.  Additional documentation on hybrid search is [here](https://docs.vectara.com/docs/learn/semantic-search/hybrid-search).
+        * Vectara can automatically provide a generative summary with the query results for Retrieval Augmented Generation (RAG).  Additional details on RAG in Vectara can be found [here](https://docs.vectara.com/docs/learn/grounded-generation/grounded-generation-overview).
+        * The Query API takes an *array* of queries, since you can have Vectara perform more than 1 query in a single API call.  The results that are returned are thus in an array as well, with each element in the response array corresponding to the input query array.
+        * Custom dimensions (`customDims`) is a feature that is only available to [Scale](https://vectara.com/pricing/) accounts.
```
tallatshafaat, line 314: The message name (aka type) is `CustomDimension`, and the field name is `dim` (i.e., `CustomDimension dim`). There is no `customDims`.

www/static/vectara-oas.yaml
```diff
@@ -224,15 +293,64 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/adminReplaceCorpusFilterAttrsRequest"
+        required: true
+      tags:
+        - AdminService
+      parameters:
+        - $ref: '#/components/parameters/customerIdentifier'
+      description: |
```
tallatshafaat, line 344: nit: curious, is there way to specify the summary of the operation? The high-level overview of this method seems weird because of being unable to use weblinks. Not important.
image

www/static/vectara-oas.yaml
```diff
@@ -287,14 +412,30 @@ paths:
...
+        * This operation authenticates with either an API Key or OAuth 2.0 (in a JWT "Bearer Token"). You can find details of how to set up an API key or use OAuth 2.0 [here](https://docs.vectara.com/docs/console-ui/manage-api-access).
+        * The `metadataJson` object needs the JSON to be escaped so that it's not confused with other parts of the request JSON.
+        * [Custom dimensions](https://docs.vectara.com/docs/learn/semantic-search/add-custom-dimensions) (`customDims`) is a feature that is only available to [Scale](https://vectara.com/pricing/) accounts.
       security:
         - ApiKeyAuth: []
         - oAuth: []
   /v1/upload:
     post:
+      description: |
+        The File Upload API can be used to index binary files like PDFs, Word Documents, and similar.
```
tallatshafaat, line 429: nit: From what I recall, this API doesn't work in the API Playground. Do you want to add a note about that?

tallatshafaat (approved):
Thanks for adding detailed descriptions to each API call. Very helpful!

I've left some minor comments. As always, "nit" means a very nitty comment and can be completely ignored.

I've tried the list-jobs and replace filters API calls from the deploy preview. Both worked fine. Thanks.

www/static/vectara-oas.yaml
eskibars: We normally use relative links, but I explicitly made these full URLs here because the links are *also* in the OpenAPI spec, which could be opened in Postman, Insomnia, etc

www/static/vectara-oas.yaml
eskibars: I was trying to figure that out, but couldn't out how yet. OpenAPI spec has a "summary" field that I expected could be used for that, but it seems to override the API name instead in this plugin for some reason. I'm going to play around with it a bit more and see if there's an update/config that helps

www/static/vectara-oas.yaml
eskibars: I'm going to try to update this PR with a new version of the plugin tomorrow. Looks like the bug might have been recently addressed: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/pull/625

eskibars: I updated the plugin with the latest version.  I *think* it will fix the file upload issue, though it's a bit difficult to test (we might need to adjust the CORS settings on the API server)

www/static/vectara-oas.yaml
eskibars: cc @pwoznic for styling comments

www/static/vectara-oas.yaml
eskibars: Disabled CORS on a new browser instance and it looks like the API does indeed work now

www/static/vectara-oas.yaml
eskibars: TY.  Pushed a fix for this

www/static/vectara-oas.yaml
eskibars: I did push an update to at least put the markdown bits below the part that shows up in this preview

www/static/vectara-oas.yaml
tallatshafaat: This is great! Thanks for fixing. https://docs.vectara.com is already is in the CORS allowed origins in prod, so hopefully this will work once deployed.

eskibars: Sorry, I should have been much clearer: it would help if we allowed CORS requests from *.netlify.app so that we can test things like this out before we release.

tallatshafaat: Ah, ok. Unfortunately, the backend code doesn't support wildcards in the allowed CORS origins. We'll have to fix that first before we can add `*.netlify.app`. Supporting wildcards will be beneficial in general, so let me create a tock-cycle issue for this.

www/docs/rest-api/vectara-rest-api.info.mdx
```diff
@@ -11,38 +11,32 @@ custom_edit_url: null
...
-## Welcome to the Vectara REST API Playground
+<h1 className={"openapi__heading"}>Vectara REST API</h1>
 
 <ApiLogo logo={{"url":"https://docs.vectara.com/img/vectara_wordmark.png","altText":"Vectara"}} darkLogo={{"url":"https://docs.vectara.com/img/vectara_wordmark_light.png","altText":"Vectara"}}></ApiLogo>
 
-Vectara is the trusted GenAI product platform. Built around a neural search 
-core, it provides an end-to-end platform for creating GenAI products using 
-a simple to use API.
+Vectara is provides an end-to-end platform for creating GenAI products using 
+a simple to use API. VectarIta is a neural search platform, built for developers
```
cjcenizal, line 25: Typo: Vectarita -> Vectara

Though honestly, I love the idea of having a Vectarita after a long day of work! 🍹

www/docs/rest-api/vectara-rest-api.info.mdx
```diff
@@ -11,38 +11,32 @@ custom_edit_url: null
...
 
-## Welcome to the Vectara REST API Playground
+<h1 className={"openapi__heading"}>Vectara REST API</h1>
 
 <ApiLogo logo={{"url":"https://docs.vectara.com/img/vectara_wordmark.png","altText":"Vectara"}} darkLogo={{"url":"https://docs.vectara.com/img/vectara_wordmark_light.png","altText":"Vectara"}}></ApiLogo>
 
-Vectara is the trusted GenAI product platform. Built around a neural search 
-core, it provides an end-to-end platform for creating GenAI products using 
-a simple to use API.
+Vectara is provides an end-to-end platform for creating GenAI products using
```
cjcenizal, line 24: Typo: "Vectara is provides" -> "Vectara is".

www/docs/rest-api/vectara-rest-api.info.mdx
cjcenizal: Actually, are both of these sentences needed? They both state "Vectara is a platform..." How about deleting the second sentence? The first one sounds clear and concise on its own to me.

www/docs/rest-api/vectara-rest-api.info.mdx
```diff
@@ -11,38 +11,32 @@ custom_edit_url: null
...
-experiment with Vectara's REST APIs directly from your browser. It offers you a 
-hands-on experience to understand and demonstrate our powerful capabilities.
-
 :::note
 
 API Keys can be used for querying and indexing operations, but cannot at this 
 time be used for administrative operations such as creating or deleting 
 corpora. These operations authenticate with a Bearer Token via the 
-OAuth 2.0 client credentials grant. Review the [**OAuth 2.0 section**](/docs/learn/authentication/oauth-2) about 
+OAuth 2.0 client credentials grant. Review the [**OAuth 2.0 section**](https://docs.vectara.com/docs/learn/authentication/oauth-2) about
```
cjcenizal, line 37: Is the domain needed? If so, should we update the other links, e.g. the one on line 29?

www/src/css/custom.css
```diff
@@ -508,6 +508,10 @@ h1
   overflow-y: initial !important;
 }
 
+.openapi-tabs__code-list-container {
+  overflow-x: scroll !important;
```
cjcenizal, line 512: How about `auto` instead of `scroll`? This will hide the scrollbar if the content doesn't overflow, and show it if the content does overflow.

cjcenizal (approved):
Didn't test locally, code LGTM. Spotted a few typos.

www/docs/rest-api/core-index.api.mdx
```diff
@@ -1,58 +1,68 @@
 ---
 id: core-index
 title: "CoreIndex"
-description: "Low level indexing API"
+description: "This API is intended to be used by experts.  It gives you fine-grained control over chunking"
```
pwoznic, line 4: The other docs section refers to low-level indexing instead of core. Should we still mention low-level anywhere in this playground topic, rename the other "Low-level API" section to Core Index (I can do in a separate PR if needed), or just leave as-is for now?

www/static/vectara-oas.yaml
pwoznic: Inside quotes is grammatically correct

www/docs/rest-api/list-jobs.api.mdx
```diff
@@ -0,0 +1,49 @@
+---
+id: list-jobs
+title: "ListJobs"
+description: "List the jobs associated with this account.  Jobs are background processes like [replacing the filterable metadata attributes](https://docs.vectara.com/docs/rest-api/replace-corpus-filter-attrs)."
```
pwoznic, line 4: Broken link (can't find replace-corpus-filter-attrs)

pwoznic (approved):
Other than the link, LGTM

www/docs/rest-api/list-jobs.api.mdx
eskibars: This is going to be fixed when we publish this documentation

PR #185, opened 2023-11-29
Update reranking.md

pwoznic:
Updated to specify the English reranker as Scale only, added the ID for the MMR, and indicated the diversity reranker is in beta. Also updated the intro and UI labels.

### DIFF
--- a/www/docs/api-reference/search-apis/reranking.md
+++ b/www/docs/api-reference/search-apis/reranking.md
@@ -26,10 +26,16 @@ default value is `10`.
 
 ## Maximal Marginal Relevance (MMR) Reranker
 
-The Maximal Marginal Relevance (MMR) Reranker enables you to diversify search 
-results to reduce redundancy while maintaining relevance to the query. You 
-enable the MMR reranker by specifying the `reranker_id` as `272725718`.
+The Maximal Marginal Relevance (MMR) reranker enables you to diversify search 
+results to reduce redundancy while maintaining relevance to the query. 
+Search queries often result in a collection of similar documents that, while 
+relevant, may lack variety. MMR addresses this by reranking the results to 
+include documents that are both relevant to your query but also different 
+from the documents already listed in the search results. This approach 
+provides users with a more balanced set of results as they may show 
+different perspectives related to your query.
 
+You enable the MMR reranker by specifying the `reranker_id` as `272725718`. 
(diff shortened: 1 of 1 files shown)


www/docs/api-reference/search-apis/reranking.md
```diff
@@ -9,45 +9,49 @@ import TabItem from '@theme/TabItem';
...
   avoiding repetition.
 * In a generative AI scenario, it produces more comprehensive summaries.
 * Diversifying results can potentially represent all points of view in the 
   data or reduce bias.
 
-In addition to specifying the `reranker_id` at query time, you also 
-specify a `diversity bias` range between `0.0` and `1.0`. Note that 
-the MMR reranker is still experimental. 
+In addition to specifying the `reranker_id` as `272725718` at query time, you also 
+specify a `diversity bias` range between `0.0` and `1.0`. Note that the diversity
```
cjcenizal, line 42: Maybe this is where we should mention "Values closer to `1.0` optimize for the most diverse results" instead of below in the context of Console? If I was crafting a query in my app code I'd expect to see this info here.

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -9,45 +9,49 @@ import TabItem from '@theme/TabItem';
...
 
-In addition to specifying the `reranker_id` at query time, you also 
-specify a `diversity bias` range between `0.0` and `1.0`. Note that 
-the MMR reranker is still experimental. 
+In addition to specifying the `reranker_id` as `272725718` at query time, you also 
+specify a `diversity bias` range between `0.0` and `1.0`. Note that the diversity 
+reranker is currently in beta. 
 
-You can also enable the MMR Reranker in the console UI as follows:
+You can also enable the beta MMR Reranker in the console UI as follows:
```
cjcenizal, line 45: Nit: I think Console should be capitalized.

cjcenizal (approved):
LGTM! Just had a couple suggestions.

pwoznic:
> Not blocking, but IMO it would be good to expand a bit on MMR beyond _what_ it does to _why_ it's an effective tool for reranking. Not super in depth, but just a bit of cursory info and at least a link to a paper/talk to provide some background for the uninitiated. I'd recommend working with @amin2718 on some details

I expanded a little on what it does and I will work with Amin on additional details and any appropriate links.

Issue #152, opened 2023-11-03
Search isn't available on narrower screens (e.g. mobile)


cjcenizal:
On narrower screens, search disappears:

![image](https://github.com/vectara/vectara-docs/assets/1238659/550c1fc5-6fec-4ffa-bb97-a39652836deb)

![image](https://github.com/vectara/vectara-docs/assets/1238659/6de1aa8f-e8c7-480e-9def-6f57f379a236)

PR #197, opened 2023-12-09
Support search on narrower screens (e.g. mobile)

cjcenizal:
Fixes https://github.com/vectara/vectara-docs/issues/152

Our current search implementation mounts an element in the nav and then injects React functionality into it to render the search component. I found it challenging to detect changes to the DOM and perform this logic in the mobile-only navigation. Once I figured that out, I realized the search component was located in an odd place in the mobile-only navigation, and couldn't figure out how to move it. This work is captured in the first commit.

While reading through the docs I found guidelines on how to implement a custom search using their `search` nav item type: https://docusaurus.io/docs/api/themes/configuration#navbar-search. I then followed https://docusaurus.io/docs/search#using-your-own-search to transplant our search component into our theme/ directory.

image


mrderyk (approved):

cjcenizal:
Per convo with @mrderyk we're going to move forward with this as-is because:
* It addresses a more important issue we have now, which is mobile-friendliness.
* It results in code that's more deeply integrated with vectara-docs, but we have 2 potential ways through which we can restore the previously-decoupled nature of the code.

We'll explore productionizing this search widget in the future, using https://docusaurus.io/docs/api/themes/@docusaurus/theme-search-algolia as a reference.

PR #202, opened 2023-12-14
Fix bug with link to vectara.com by using Docusaurus Link component.

cjcenizal:


www/src/theme/search/SearchModal.tsx
```diff
@@ -10,14 +10,14 @@ import { SearchInput } from "./SearchInput";
...
   VuiFlexItem,
-  VuiLinkInternal,
   VuiPortal,
   VuiScreenBlock,
   VuiSpacer,
   VuiText,
   VuiTextColor,
 } from "./vui";
 import { FocusOn } from "react-focus-on";
+import Link from "@docusaurus/Link";
```
cjcenizal, line 20: @mrderyk Note the tight-coupling with Docusaurus. Something to consider when we generalize this search widget.

mrderyk (approved):

Issue #190, opened 2023-12-01
IndexDocument API docs are out of date


amin2718:
### Pages

https://docs.vectara.com/docs/api-reference/indexing-apis/indexing

### Description

![image](https://github.com/vectara/vectara-docs/assets/83042956/b5400a2c-35a7-42ca-9e35-c14c0a25fa34)

These assertions were true until June 2022, but are no longer true. The current definition is:

```
message IndexDocumentResponse {
  // If ALREADY_EXISTS, it means the document was already indexed, and no new quota was consumed.
  Status status = 1;

  // The storage quota needed for the document indexed in the request.
  // If "status" is ALREADY_EXISTS, it means that the document was already in the index prior to
  // this request. In such cases, quota is not consumed again and the value in this field
  // represents the quota consumed when the document was indexed the first time.
  StorageQuota quota_consumed = 2;
}
```

Also, the docs don't account for instant indexing. In most cases, the data is available for query serving in <1 second. When the documentation was written, the delay was up to 5 minutes.

amin2718:
I agree that it's hard to keep the snippets updated all the time. Given the underlying system is React, it should be possible for Deryk or someone to create a custom tag that pulls the definition of any protobuf into a nicely formatted code listing:

```
<ProtobufListing name="indexing.IndexDocumentResponse>
```

Barring some kind of effort like that, we'll have to remove snippets from page. That does have the downside that users end up jumping back and forth between our long-form documentation and some other source code browser to fully understand what's going on.

Anyway, your call on the code snippets. In any case, the narrative text that explains how to make the indexing request and interpret the response is also incorrect.

amin2718:
Here's a concrete example where I can't point a customer to our docs to help diagnose the root cause of an indexing issue, because this document is out of date.

![image](https://github.com/vectara/vectara-docs/assets/83042956/03f4e67c-efb4-4f8b-a762-965a98dd1f2c)

What I'd like to ask them is to report to me the breakdown of text vs metadata when indexing the problematic pdf.

amin2718:
This came up again today from a different potential customer. Play.ht. They were asking whether there is any way to know how much space an indexed document is taking up. I had to explain that our index API returns this information, but the documentation is simply not up to date.

eskibars:
@pwoznic I think this can be closed, since the gRPC definitions were pulled out of the docs.  Is that accurate?

PR #205, opened 2023-12-17
Addressed #190: Indexing documentation is out of date.

amin2718:
While working on Issue #190 , I also fixed:

1. Inaccurate information about how long it generally takes information that's indexed to come online. In the era of instant indexing, it's generally just a second.
2. Delete document endpoint information was completely missing, so I added it.

### DIFF
--- a/www/docs/api-reference/indexing-apis/indexing.md
+++ b/www/docs/api-reference/indexing-apis/indexing.md
@@ -77,7 +77,8 @@ index. The Index request requires the following parameters:
 ```
 
 The response from the server includes a status code and the amount of quota
-consumed. For details, please see
+consumed. For details, please see [IndexDocument](#index-document).
+
 
 Let's take a closer look at the document object which encapsulates the
 information about the document to be indexed. It typically includes the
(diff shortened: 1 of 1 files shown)


www/docs/api-reference/indexing-apis/indexing.md
```diff
@@ -75,10 +75,14 @@ index. The Index request requires the following parameters:
   }
 }
 ```
-Let's take a closer look at the document object which encapsulates the 
-information about the document to be indexed. It typically includes the 
-title, description, and metadata. The core of the document is also structured 
-in sections that can include unique identifiers, titles, strings, metadata, 
+
+The response from the server includes a status code and the amount of quota
+consumed. For details, please see
```
pwoznic, line 80: Was this meant to be a link to another topic or leading into sample syntax? FYI, I wanted this topic to also have REST examples

amin2718: Thanks for catching that. It's fixed now.
As far as REST examples, that's a good idea. We can leave the bug open until those are added. I needed to put at least the gRPC in because it's caused some issues in two customer discussions recently.

pwoznic (approved):
LGTM

PR #214, opened 2023-12-29
Add Content-Disposition header to set the Document ID

### LINKED ISSUE #210: File Upload API Definition missing information about how the document id is set
### Idea

https://docs.vectara.com/docs/api-reference/indexing-apis/file-upload/file-upload

The file upload API documentation doesn't provide any guidance on how to set the document id in our system. This is a serious omission given the centrality of the id.

The id is taken from the `Content-Disposition` header. The value corresponding to the `filename` key is extracted.

Reference: https://bitbucket.org/zir-ai/platform/src/aaca4a16b19d55a35325318f3a04efb919febfec/indexing/titan/src/main/java/com/vectara/indexer/baobab/FileUploadServlet.java#lines-659

Context: Engineers at play.ht were confused and couldn't find guidance in our docs.

pwoznic:
Fixes #210 

Adding this header lets you set the document ID when using the file upload API


aamirbutt (approved):
Although this statement is correct. But Content-Disposition's primary purpose is to specify the `form-data`
We use the `filename` as the ID which is something specific to our platform. 

I would suggest adding an outside link to Content Disposition header as well, so users know what it is. For example, following link seems good. 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

PR #216, opened 2023-12-30
Update part.is_title field description

### LINKED ISSUE #212: The explanation of how is_title works is confusing
### Idea

When trying to understand how is_title works - the search leads me to this page: https://docs.vectara.com/docs/learn/metadata-search-filtering/ootb-metadata-filters

This only mentions that part.is_title is a default field, but I think it's worthwhile to better explain how titles are handled in our system more broadly. From my understanding:
1. During indexing: how are document-level and section-level titles indexed?
2. How are titles used at query time if no filter is specified? if using a filter condition of "is_title.part=true" and if using a filter condition of "is_title.part<>true"
3. Are titles included in matching during query-time? 
4. General guidance on how/when to use is_title as a filter.

pwoznic:
Fixes #212 

Updated part.is_title field to include when it's not equal to true (<>). This is useful when documents do not have titles and you want to search for sections without titles. Setting to false will not work because there are cases where this field is unset.

### DIFF
--- a/www/docs/learn/metadata-search-filtering/ootb-metadata-filters.md
+++ b/www/docs/learn/metadata-search-filtering/ootb-metadata-filters.md
@@ -5,8 +5,9 @@ title: Default Metadata Filters
 
 import {Config} from '@site/docs/definitions.md';
 
-A few pieces of metadata expressions filterable out of the box, as they're very
-useful in a variety of situations.
+A few pieces of metadata expressions are filterable out of the box, including 
+Document ID, Language, and Titles. These filters are very useful in a variety 
+of situations.
 
 Note that you can set up additional fields to filter on by setting up
 [filter attributes](/docs/api-reference/admin-apis/create-corpus#filter-attribute) on a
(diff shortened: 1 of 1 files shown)


www/docs/learn/metadata-search-filtering/ootb-metadata-filters.md
```diff
@@ -37,12 +40,19 @@ Valid filter expressions for this would be something like:
...
 
-To filter for title fields only, you can use: `part.is_title = true` and
-conversely `part.is_title = false` will return only non-title sections.
+This field actually uses three value logic: true, false, and unset. We 
+designed it like this to avoid creating too much metadata because customers 
+are billed for metadata, so doing so is in the customer's interest. Here is 
+how it works:
 
+* To filter for title fields only, use `part.is_title = true`
+* To return only non-title sections, use `part.is_title = false`
```
ofermend, line 56: I'm not sure I understand how `part.is_title = false` works.
Perhaps it'd be useful to clarify with an example document that is indexed to show what is included or not included with each case?

www/docs/learn/metadata-search-filtering/ootb-metadata-filters.md
```diff
@@ -50,9 +50,14 @@ in which case this field can be used to filter.
 This field actually uses three value logic: true, false, and unset. We 
 designed it like this to avoid creating too much metadata because customers 
 are billed for metadata, so doing so is in the customer's interest. Here is
```
ofermend, line 52: "so doing so is in..." - too many uses of "so" maybe suggest to rephrase?

Issue #223, opened 2024-01-11
Docs should describe how to locate the customer ID somewhere


cjcenizal:
Many API requests need to find the customer ID but when I searched the docs for "customer ID" I couldn't find any helpful matches. It would be helpful if the docs described where to find the customer ID in the Console UI.

PR #224, opened 2024-01-12
Update console and corpus topics

### LINKED ISSUE #222: Create Corpus docs needs to explicitly mention "corpus ID"
The corpus ID is used in many API requests, and folks will visit the Console UI to find this ID. I suggest updating https://docs.vectara.com/docs/console-ui/creating-a-corpus to mention where to find the corpus ID, e.g.

> Vectara API requests against a corpus require the corpus ID. Find the corpus ID in the top-left corner of the corpus view, near the corpus name.

pwoznic:
Fixes #222 and #223


cjcenizal (approved):
Thanks for making these changes!

PR #230, opened 2024-01-25
Add Create-UI and refine copy and layout of Build Apps landing page

cjcenizal:


www/docs/build-apps/app-building.md
```diff
@@ -7,28 +7,35 @@ sidebar_label: App Building Tools
...
-  data.
-## [React-Search](/docs/build-apps/react-search)
-  
-  Seamlessly integrate Vectara’s semantic search capabilities into your React 
-  applications with just a few lines of code.
+## Data ingestion
 
+### [Vectara Ingest](/docs/build-apps/vectara-ingest)
+
+Sample templates for pulling data from many popular data sources.
```
mrderyk, line 19: I'm not sure this fully encompasses what vectara-ingest is. The templates themselves are only configuration files, but vectara-ingest also includes [the various crawlers](https://github.com/vectara/vectara-ingest/tree/main/crawlers) that actually consume those configurations.

www/docs/build-apps/react-search.md
```diff
@@ -7,9 +7,11 @@ sidebar_label: React-Search
 import {vars} from '@site/static/variables.json';
 import {Config} from '@site/docs/definitions.md';
 
-The Reach-Search component pulls data from your Vectara account and adds 
-semantic search to your React apps with a few lines of code. [Interact with the 
+![Screenshot of React-Search search box](/img/react_search.jpg)
+
+The Reach-Search component pulls data from your Vectara account and adds
```
mrderyk, line 12: typo: Reach -> React

mrderyk, line 12: Do we want to say account or use corpora/datastore instead, for consistency of terminology?

mrderyk (changes requested):
One typo and some non-blocking notes. :)

www/docs/build-apps/app-building.md
cjcenizal: You're right. I was thinking of the crawlers as a form of template, too. Or at least it would be implied that they're included, otherwise what would you do with the templates? Do you think folks will be confused?

cjcenizal: Never mind, better to be safe than sorry. :) Incorporated your suggestion.

cjcenizal:
@mrderyk I addressed your comments. Can you take another look?

mrderyk (approved):

PR #238, opened 2024-02-20
Add Generative Prompts section

pwoznic:
Added new prompts section including topic about customizing prompts with metadata.

### DIFF
--- a/www/docs/prompts/custom_prompts_with_metadata.md
+++ b/www/docs/prompts/custom_prompts_with_metadata.md
@@ -26,7 +26,7 @@ The following table shows the available custom prompt variables:
 | $vectaraQuery  | The query provided by the user  | Generate a summary in $vectaraOutChars characters in language '${vectaraLangCode}' for the query \"$esc.java(${vectaraQuery})\" solely based on the search results in this chat.  | Generate a summary in 512 characters in language 'ara' for the query \"Give me \"some\" search results.\" solely based on the search results in this chat.  |
 | $vectaraIdxWord  | A utility array to convert the index to words i.e "first", "second", "third", "forth", "fifth", "sixth", "seventh", "eighth", "ninth", "tenth"  | $vectaraIdxWord[0]  | first  |
 | $vectaraLangName  | Set to the requested language name. The language can either be requested explicitly or detected from the language of the query.  | You are a helpful assistant. Answer in ${vectaraLangName}.  | You are a helpful assistant. Answer in Arabic.  |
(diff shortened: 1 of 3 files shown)


www/docs/prompts/custom_prompts_with_metadata.md
```diff
@@ -0,0 +1,114 @@
...
+| $vectaraQuery  | The query provided by the user  | Generate a summary in $vectaraOutChars characters in language '${vectaraLangCode}' for the query \"$esc.java(${vectaraQuery})\" solely based on the search results in this chat.  | Generate a summary in 512 characters in language 'ara' for the query \"Give me \"some\" search results.\" solely based on the search results in this chat.  |
+| $vectaraIdxWord  | A utility array to convert the index to words i.e "first", "second", "third", "forth", "fifth", "sixth", "seventh", "eighth", "ninth", "tenth"  | $vectaraIdxWord[0]  | first  |
+| $vectaraLangName  | Set to the requested language name. The language can either be requested explicitly or detected from the language of the query.  | You are a helpful assistant. Answer in ${vectaraLangName}.  | You are a helpful assistant. Answer in Arabic.  |
+| $vectaraQueryResults  | An array of query results is found in the response, sorted by relevance score.  | You are a helpful assistant. Answer in ${vectaraLangName}.  | {"role": "user", "content": "Give me the second search result."},{"role": "assistant", "content": "2nd result" },  |
```
eskibars, line 29: @pwoznic the Usage Input and Usage Output are not correct here.
The example input would be something like: `#foreach $qResult in $vectaraQueryResults`

The example output would be something like:`[ result_array ]`, but it's a bit difficult to show the result array in a meaningful way beyond that.

www/docs/prompts/custom_prompts_with_metadata.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Example Custom Prompt for an RFI Answering Bot
+
+The following example prompt creates a Request for information (RFI) 
+answering bot that includes metadata. First, we ask the generative LLM to 
+answer an RFI question and tell it how the results will come back from the 
+query. 
+
+We want to iterate through `$vectaraQueryResults` inserting the results 
+in the order that we like. `qResult.getText()` provides the most relelvant
```
eskibars, line 71: `$qResult` instead of `qResult`

eskibars (approved):
A few small changes to address; I've left suggestions.  Go ahead and merge after those are made

www/docs/prompts/custom_prompts_with_metadata.md
pwoznic: Updated

www/docs/prompts/custom_prompts_with_metadata.md
pwoznic: Updated example

PR #241, opened 2024-02-23
Add new Scale features to the playground

eskibars:
Adds `ModelParams`, `debug` and `responseChars` -- all Scale features -- to the API playground.  @pwoznic we should also add these to some of the long-form docs for generative requests


pwoznic (approved):
LGTM, I will create a separate PR for long-form updates

PR #245, opened 2024-02-29
fix: Only execute analytics page function if it is available.

mrderyk:
## CONTEXT

The window.analytics.page function needs to execute when the user navigates to a new page in the docs site. However, when the user navigates to a specific page (i.e. deeplink), the function attempts to fire before it is defined.

## CHANGE
- Use optional chaining to only execute `window.analytics.page` if it is defined.

### DIFF
--- a/www/static/routeUpdateModule.js
+++ b/www/static/routeUpdateModule.js
@@ -1,10 +1,7 @@
-
 export function onRouteDidUpdate({ location, previousLocation }) {
   // Don't execute if we are still on the same page; the lifecycle may be fired
   // because the hash changes (e.g. when navigating between headings)
   if (location.pathname !== previousLocation?.pathname) {
-    if (window.analytics) {
-      window.analytics.page();
-    }
+    window.analytics?.page?.();
   }
-};
+}


www/static/routeUpdateModule.js
```diff
@@ -1,10 +1,7 @@
-
 export function onRouteDidUpdate({ location, previousLocation }) {
   // Don't execute if we are still on the same page; the lifecycle may be fired
   // because the hash changes (e.g. when navigating between headings)
   if (location.pathname !== previousLocation?.pathname) {
-    if (window.analytics) {
-      window.analytics.page();
-    }
+    window.analytics?.page?.();
```
cjcenizal, line 5: Am I correct in assuming there are scenarios where we don't get analytics because the object isn't available yet? In that case, do we want to retry until it is available?

cjcenizal (commented):
> however, when the user navigates to a specific page (i.e. deeplink)

Can you give an example of this? If this like if I sent you a link in Slack to https://docs.vectara.com/docs/api-recipes and you clicked it, that's considered a deeplink in this scenario?

cjcenizal (approved):
Approving to unblock this so we can fix the docs. We can address my questions separately.

www/static/routeUpdateModule.js
mrderyk: I thought of that, but when I first commented this out to try diagnosing the issue, I saw that the ui_event was being triggered anyway.

My theory is that an event triggers:
a) from a fresh nav (i.e. via deeplink, navigating to the homepage), and 
b) whatever's in routeUpdateModule.js gets triggered on a route change, i.e. a client side refresh

For some reason, if we directly navigate to something other than the homepage, case a) occurs, and case b) attempts to happen, but the function isn't defined.

PR #252, opened 2024-03-23
Add information about the Factual Consistency Score

pwoznic:
Updated the following:

- Added FCS information to topics
- Created a separate topic for the Query tab of a corpus with the new search app types
- Updated screenshots


www/docs/console-ui/corpus_query_configuration.md
```diff
@@ -0,0 +1,88 @@
+---
```
cjcenizal, line 1: Who is this page for and why will they need it? I'm having trouble imagining someone coming here to learn about the Console UI. Wouldn't they just use the Console UI? I'm also concerned about maintenance. This is a lot of content to keep in sync with the UI.

I suggest rethinking this page from the perspective of someone who has a problem to solve. What is that problem, and how does this page solve it?

Example problems:
* "Can I use Vectara to chat/search/summarize and how do I do that?"
* "What are the differences between chat, search, and summarization?"

www/docs/learn/grounded-generation/grounded-generation-overview.md
```diff
@@ -244,6 +244,47 @@ Hitchhiker's Guide to the Galaxy:
...
+          "response_lang": "en",
+          "factual_consistency_score": true
+          },
+        }
+      ]
+```
+
+The Vectara Factual Consistency Score automatically evaluates and detects 
+hallucinations in generated output. This score can range from `0.0` to `1.0`. 
+Higher scores indicate a greater probability of being factually accurate,
```
cjcenizal, line 269: We switch from "score" to "scores" in this sentence. Should we retain the pluralization of the subject?

cjcenizal (approved):
LGTM but I can't speak to the accuracy of the bits about the factual consistency score.

www/docs/console-ui/corpus_query_configuration.md
pwoznic: Thanks, CJ! The Query tab has been updated with a lot of great stuff lately and I felt like it could use its own topic. The topic may be too busy now and not really answer what are the differences, so I can reduce the amount of screenshots and give more of an overview on these points. The intent was to showcase the various drawers and configuration options, search app types, etc for users who may come across this console documentation before using the product (e.g. they found it on Google when searching for something like semantic search).

www/docs/learn/grounded-generation/grounded-generation-overview.md
pwoznic: I went back to singular in the latest commit.

pwoznic:
I revamped this topic to provide more context

PR #256, opened 2024-04-02
Add Stream Query API Definition topic

pwoznic:
Added Stream Query long-form documentation topic.

### DIFF
--- a/www/docs/api-reference/search-apis/stream-query.md
+++ b/www/docs/api-reference/search-apis/stream-query.md
@@ -1,7 +1,7 @@
 ---
 id: stream-query
 title: Stream Query API Definition
-sidebar_label: StreamQuery API Definition
+sidebar_label: Stream Query API Definition
 ---
 
 import Tabs from '@theme/Tabs';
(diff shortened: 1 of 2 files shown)


www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,99 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import {Config} from '@site/docs/definitions.md';
+import {vars} from '@site/static/variables.json';
+
+The Stream Query API enables continuous streamed responses as data becomes 
+available, improving responsiveness and reducing latency. Instead of waiting 
+for the entire response to complete, you get query results and summaries in a
```
cjcenizal, line 14: > Instead of waiting for the entire response to complete, you get query results and summaries in a streaming manner. 

I think it will help the reader grok this feature if we add some more details like this:

> Instead of receiving a complete response, consumers will receive partial responses in this order:
> 1. Search results.
> 2. If summarization is enabled, chunks of the summary, e.g. "This", "is", "a", "summary".
> 3. If the factual consistency score (FCS) is enabled, then the FCS will be the final response.
>
> Note that this API only behaves differently than the [Standard Query API](/docs/api-reference/search-apis/search) if summarization is enabled.

www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,99 @@
...
+The Stream Query API enables continuous streamed responses as data becomes 
+available, improving responsiveness and reducing latency. Instead of waiting 
+for the entire response to complete, you get query results and summaries in a 
+streaming manner. The Stream Query API streams the response back in smaller 
+chunks as the data is generated, rather than returning the complete response 
+in a single message like the [Standard Query API](/docs/api-reference/search-apis/search).
+
+This streaming approach is beneficial when generating summaries using LLMs 
+like GPT-4, which can have significant latencies of 5-10 seconds. 
+In contrast, the Standard Query API makes users wait for the full summarization
```
cjcenizal, line 21: I think this last bit repeats ideas already mentioned earlier so maybe we can edit it out? WDYT?

> In contrast, the Standard Query API makes users wait for the full summarization 
process before receiving any results. With streaming, the summary request 
processes with near-zero latency, significantly enhancing the user experience.

www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,99 @@
...
+This streaming approach is beneficial when generating summaries using LLMs 
+like GPT-4, which can have significant latencies of 5-10 seconds. 
+In contrast, the Standard Query API makes users wait for the full summarization 
+process before receiving any results. With streaming, the summary request 
+processes with near-zero latency, significantly enhancing the user experience.
+
+
+## Stream Query Request Body
+
+The Stream Query API has the same request parameters as the [Standard Query API](/docs/api-reference/search-apis/search).
```
cjcenizal, line 28: We can improve the DX by duplicating the parameters here instead of making the reader open another tab.

www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,99 @@
...
+
+Each streamed chunk contains a portion of the summary text, identified by a 
+unique `future_id`. Once the full summary is streamed, you receive a final 
+response with the `done` field set to `true`, allowing flexible handling and 
+processing of results. The Stream Query API request has three different types 
+of responses:
+
+### Preamble Response
+
+This initial response establishes `future_id` values and serves as a preamble,
```
cjcenizal, line 43: I don't really understand how future IDs are supposed to be used. Maybe an example would help?

www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Stream Query Request Body
+
+The Stream Query API has the same request parameters as the [Standard Query API](/docs/api-reference/search-apis/search).
+The `stream-query` endpoint enables streaming. Use this endpoint instead of 
+the standard `query` endpoint.
+
+
+## Stream Query Response Types
```
cjcenizal, line 33: Should we also mention that FCS is part of the response, and it comes after the summary has `done: true` set?

www/docs/api-reference/search-apis/stream-query.md
ofermend, line 21: Suggest removing "In contrast" here and add it before the streaming (next sentence)

www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Streamed Parts of the Summary Response
+
+The third response type, which streams until you get the final `done` value, 
+returns the subsequent streamed chunks of the summary. Each response has a 
+`batchQueryResponse` that contains a portion of the `summary` text. The 
+`future_id` value in this response matches the `future_id` assigned to the 
+search results of the initial response.
+
+## Concatenation of Streamed Data Chunks
```
cjcenizal, line 64: Hmmm this doesn't reflect how we're consuming chunks in Console or our OSS projects. I'll check in with Tallat and circle back.

cjcenizal (commented):
Thanks so much for putting this together so quickly! I'm not ready to approve yet because the stuff about concatenation is all new to me so I need to dig into it a bit with the other engineers.

www/docs/api-reference/search-apis/stream-query.md
pwoznic: Thanks, CJ. This was my attempt at putting something together and trying to keep it somewhat generic since the consuming app is responsible for forming the response. I'll leave it to the SMEs to correct me! :)

cjcenizal, line 64: I think we should remove this section entirely for now and also remove all mentions of `future_id` for now. This section doesn't seem correct based on my conversation with Amin and I think we need more clarity on the use cases for `future_id`.

I suggest we defer two scopes of work to unblock this from being published:
* We'll add a concatenation explanation and example.
* We'll add a thorough description of the meaning and use cases for `future_id`.

cjcenizal (changes requested):

www/docs/api-reference/search-apis/stream-query.md
```diff
@@ -0,0 +1,217 @@
...
+### Search Results Response
+
+This second response type contains the search results as the 
+`batchQueryResponse` populates with these results in real time.
+
+### Streamed Parts of the Summary Response
+
+The third response type, which streams until you get the final `done` value, 
+returns the subsequent streamed chunks of the summary. Each response has a 
+`batchQueryResponse` that contains a portion of the `summary` text
```
cjcenizal, line 79: Missing period at end of sentence.

cjcenizal (approved):
Looks good! Had one comment.

PR #259, opened 2024-04-03
Update Information Architecture with new sections for Capabilities and Use Cases

pwoznic:
Updated the IA to split the Learn section which has become too large and align with current industry best practices:

- Moved the Data Privacy and Authentication content to its own section
- Created a Capabilities section to cover the main concepts such as Semantic Search, RAG, and so on
- Created a Use Cases and Guides section for procedures and tutorials. The intent is to make these procedures easier to find via search or TOC


eskibars:
Nice!  This is much better already.

Some comments/suggestions:
- "Data Privacy and Authentication" is confusing since there's a separate top level section "API Authentication Examples".  As a user, I'd question "which do I go to for Authentication?"  I'd personally rename this section to "Security and Data Privacy" (security and privacy are intertwined, and security feels more holistic than just auth) and drop the API Auth Examples up into that section as a sub-section
- Related to the previous: the verbiage isn't then consistent in other parts of the doc.  That is, in the API reference, Authentication is called "Access" which was unexpected to me (I wasn't sure what would be in "Access").  Also, all of these "Access" (Authentication) APIs don't look like they're cross-linked from the previous top-level section as far as I could tell, so it makes them even less discoverable
- Some of the sections have verb prefixes while others don't, and of different tenses and different verbs.  I'm pretty sure I'm responsible for some of these, but for example "Enable Textless Mode" (present simple) "Keeping Your Data Private" (present continuous), all of the Use Cases and Guides use a present simple verb but none in the "Capabilities" section.  Lots of synonyms ("Enable," "Add," Keep," "Select," "Configure," etc) for the purposes of this type of documentation also makes it more difficult to read. Let's be consistent: I'd recommend removing all of the verb prefixes which will have a 2nd advantage of keeping the sidebars narrower and wrap multiline less
- I'd create some specific structure around what goes into the "Use Cases and Guides" -- are they individual features/capabilities in Vectara or are they broader use cases?  Here, they're a mix ("Add Custom Dimensions to Enhance Scoring" is an objective based title "...to enhance scoring" even though it's really only about Custom Dimensions, "Recommendation System" is _mostly_ about using `RESPONSE` semantics and doesn't really go into depth on how to actually build/use a Recommendation System with Vectara beyond that specific feature, the Pagination page is really just about that feature of Vectara as opposed to a use case/guide).  Some of these feel like they're really just Capabilities of the platform (e.g. pagination and custom dimensions).  It especially feels confusing because most of the items that now live in the Capabilities section feel more like Use Cases/Guides (e.g. RAG, document data structuring, the "Semantic Search" section).  The last of those has a nice feature that it starts off with a "Fundamentals" section and then provides 1 specific guide on relevance tuning that actually touches on several capabilities.  My recommendation would be to write down some guidance on e.g. a notion page of what the structure of each of these is.  I'd recommend Use Cases and Guides would typically be use cases or a broad guide like the document data structuring guide that starts off with an overview page like the Semantic Search section in Capabilities and then dives into specifics of how to configure the composite pieces ("Capabilities") for the use case or objective.  Capabilities then would describe a much smaller feature/aspect of the APIs (e.g. pagination or HHEM)
- Generative prompts stands out for 3 reasons:
  - It's a top-level nav item when it's really "just" a capability of the system.  It's not clear to me why it gets its own top-level item
  - It's not linked to anywhere from the "RAG" capability (which is really more like a Use Case guide)
  - There are separate parts of the generative capabilities that are spread out like "Configure Response Languages" (which feels to me more like just a capability of the generative controls) and "Select a Summarizer" which has both a combination of some "Capabilities-like" stuff _and_ a little bit of a use case guide

pwoznic:
> Some comments/suggestions:
> 
> * "Data Privacy and Authentication" is confusing since there's a separate top level section "API Authentication Examples".  As a user, I'd question "which do I go to for Authentication?"  I'd personally rename this section to "Security and Data Privacy" (security and privacy are intertwined, and security feels more holistic than just auth) and drop the API Auth Examples up into that section as a sub-section
> * Related to the previous: the verbiage isn't then consistent in other parts of the doc.  That is, in the API reference, Authentication is called "Access" which was unexpected to me (I wasn't sure what would be in "Access").  Also, all of these "Access" (Authentication) APIs don't look like they're cross-linked from the previous top-level section as far as I could tell, so it makes them even less discoverable
> * Some of the sections have verb prefixes while others don't, and of different tenses and different verbs.  I'm pretty sure I'm responsible for some of these, but for example "Enable Textless Mode" (present simple) "Keeping Your Data Private" (present continuous), all of the Use Cases and Guides use a present simple verb but none in the "Capabilities" section.  Lots of synonyms ("Enable," "Add," Keep," "Select," "Configure," etc) for the purposes of this type of documentation also makes it more difficult to read. Let's be consistent: I'd recommend removing all of the verb prefixes which will have a 2nd advantage of keeping the sidebars narrower and wrap multiline less
> * I'd create some specific structure around what goes into the "Use Cases and Guides" -- are they individual features/capabilities in Vectara or are they broader use cases?  Here, they're a mix ("Add Custom Dimensions to Enhance Scoring" is an objective based title "...to enhance scoring" even though it's really only about Custom Dimensions, "Recommendation System" is _mostly_ about using `RESPONSE` semantics and doesn't really go into depth on how to actually build/use a Recommendation System with Vectara beyond that specific feature, the Pagination page is really just about that feature of Vectara as opposed to a use case/guide).  Some of these feel like they're really just Capabilities of the platform (e.g. pagination and custom dimensions).  It especially feels confusing because most of the items that now live in the Capabilities section feel more like Use Cases/Guides (e.g. RAG, document data structuring, the "Semantic Search" section).  The last of those has a nice feature that it starts off with a "Fundamentals" section and then provides 1 specific guide on relevance tuning that actually touches on several capabilities.  My recommendation would be to write down some guidance on e.g. a notion page of what the structure of each of these is.  I'd recommend Use Cases and Guides would typically be use cases or a broad guide like the document data structuring guide that starts off with an overview page like the Semantic Search section in Capabilities and then dives into specifics of how to configure the composite pieces ("Capabilities") for the use case or objective.  Capabilities then would describe a much smaller feature/aspect of the APIs (e.g. pagination or HHEM)
> * Generative prompts stands out for 3 reasons:
>   
>   * It's a top-level nav item when it's really "just" a capability of the system.  It's not clear to me why it gets its own top-level item
>   * It's not linked to anywhere from the "RAG" capability (which is really more like a Use Case guide)
>   * There are separate parts of the generative capabilities that are spread out like "Configure Response Languages" (which feels to me more like just a capability of the generative controls) and "Select a Summarizer" which has both a combination of some "Capabilities-like" stuff _and_ a little bit of a use case guide
> * Related to the previous bullet, https://deploy-preview-259--luxury-shortbread-acee05.netlify.app/docs/learn/grounded-generation/select-a-summarizer#add-a-timeout-to-rest-calls is more like a general capability of the APIs across the board: nothing specific to summarizers.  Its just that you're more likely to bump into this with summarization because summarization can take a longer tim

Thanks I'll start a notion an reassess everything

pwoznic:
Fixes #271

PR #262, opened 2024-04-16
Update Create Corpus default example

pwoznic:
Updated the Create Corpus default example to be more simplified.


ofermend (approved):
LGTM

PR #264, opened 2024-04-17
Update prompts to use single-quote in examples

pwoznic:
Updated examples in the prompt engine and custom prompts topics


ofermend (approved):
LGTM!

Issue #269, opened 2024-04-25
Add summarization to docs search


cjcenizal:
Per convo with @eskibars let's surface a summarized response in the search widget, to help folks find answers to their questions better.

PR #274, opened 2024-05-07
Add summarized search

cjcenizal:
Fixes https://github.com/vectara/vectara-docs/issues/269

Changes:
* Migrate to React-Search dependency.
* Enable summarized search.
* Handle change to summary toggle, persist it in local storage, and then use it to set the initial toggle state.

image


www/src/theme/search/Search.tsx
mrderyk, line 1: This cleanup 😍

www/src/theme/SearchBar.tsx
```diff
@@ -1,21 +1,35 @@
...
 const CUSTOMER_ID = "3874164736";
 const CORPUS_ID = "1";
 const API_KEY = "zqt_5usQAFwTytdQHAXn17Iq31OQMA5RrIBWLI7Fwg";
 
+const persistToggleSummaryState = (isSummaryVisible) => {
+  localStorage.setItem("isSummaryVisible", isSummaryVisible ? "true" : "false");
+};
+
+const isSummaryToggleInitiallyEnabled =
+  localStorage.getItem("isSummaryVisible");
```
mrderyk, line 14: Do you think we would want to cache the user choice for other places we have search as well (right now, just console, but we could have more later)? If so, should we namespace this so that naming localStorage keys in the future becomes more uniform (e.g. "docs-search-isSummaryVisible" and "console-search-isSummaryVisible")?

mrderyk (approved):
Looks good. Just had one question regarding local storage keys.

Also, not sure why every build step seems to be failing.

cjcenizal:
@mrderyk Thanks for pointing out those build errors! I fixed them. They were due to React-Search extending `HTMLElement` when it defines its custom webcomponent. Docusaurus's build step assumes a Node environment (for SSR), and `HTMLElement` only exists in the browser, so it caused an `undefined` reference error. For future reference, the solution was to use some of the tools described in https://docusaurus.io/docs/docusaurus-core for isolating browser-specific logic.

PR #276, opened 2024-05-08
Update custom prompt topics with Query endpoint details and links

### LINKED ISSUE #273: Custom prompt docs need more context
This is in relation to:
* https://docs.vectara.com/docs/prompts/vectara-prompt-engine
* https://docs.vectara.com/docs/prompts/custom-prompts-with-metadata

It's not clear which API these docs are connected to. How do we use the objects described in these docs? Which field are they assigned to? How does this relate to the `promptText` field?

pwoznic:
Fixes #273 

Updated topics in the Generative prompts section to cross-link to the Query endpoint topics and custom prompt example in the API Playground


cjcenizal:
Thanks for making these changes @pwoznic! But I think my original concerns remain. Let me try to unpack the confusion I encounter with each page and suggest some changes.

## Vectara Prompt Engine

When I read this page, my introduction into the topic is a description of the role this component plays within our platform. This is good information but it's role as the introduction has two undesirable effects on me:
* It presents the information from Vectara's point-of-view, which is on the opposite side of _my_ point-of-view. So I need to try to assume the point-of-view of the platform, which is unusual and uncertain. This slows me down and confuses me about why I'm reading about this stuff -- I'm here to solve _my_ problem, so if something isn't presented from my point-of-view, is it even for me?
* It makes me think about Vectara's internals, which is counter to our core value prop: that the Vectara platform encapsulates complexity for you and things generally just work with minimal fuss. I think we're counter-messaging that value prop by exposing some internals in the introduction

My suggestion would be to rename this page "Custom prompts" and to begin with an explanation of `promptText` -- essentially elevating the callout you added to become the introductory paragraph. Then add a section called "Prompt templates" though to be honest I'm not sure if that's a thing. Are they a thing? I'm guessing prompt templates are a thing that is then consumed by the Vectara Prompt Engine? If that's the case, then I would nest "Example prompt template", "Effective prompts", and "Prompt design" in that section, in that order, because examples are a great way to show (not tell) and also provide people with something concrete to anchor to as they read a description of the thing. I then suggest adding a another sub-section called "Vectara Prompt Engine" at the very bottom of the "Prompt templates" section, and this is where I'd describe the engine itself, framed as "All of this functionality is powered by our Prompt Engine..." It's the last thing I want to read about, because it's the least actionable piece of information.

## Custom Prompts with Metadata

When I read this page, I don't understand the context at all. Are these variables used with prompt templates? Custom prompts? Can we explicitly describe how/where people should use them? Is the "Custom prompts" section I described required reading to understand this page? If so, then let's state that and link back to it.

Hope this is helpful!

PR #279, opened 2024-05-20
Update citation section

pwoznic:
Updated the citation section based on discussion such as adding the new summarizers for citations.


ofermend (approved):
LGTM

PR #280, opened 2024-05-20
chore: Point ReactSearch to updated reranker

mrderyk:
## CONTEXT

We need to configure docs search to use the latest reranker

## CHANGES
- Add rerank configuration to ReactSearch
- Point ReactSearch to docs corpus used for demos.


cjcenizal (approved):
Code LGTM, didn't test locally.

PR #293, opened 2024-06-12
Change nav from API Playground -> API Reference.

cjcenizal:
This is a more accurate description of this section, now that we have a comprehensive spec. 
image

### DIFF
--- a/www/docusaurus.config.js
+++ b/www/docusaurus.config.js
@@ -1,11 +1,11 @@
 // @ts-check
 // Note: type annotations allow type checking and IDEs autocompletion
 
-
 /** @type {import('@docusaurus/types').Config} */
 const config = {
   title: "Vectara Docs",
-  tagline: "Developer documentation for Vectara's Retrieval Augmented Generation as-a-Service (RAGaaS) Platform",
+  tagline:
+    "Developer documentation for Vectara's Retrieval Augmented Generation as-a-Service (RAGaaS) Platform",
   url: "https://docs.vectara.com",
   baseUrl: "/",
   favicon: "img/vectara_logo.svg",
@@ -57,12 +57,12 @@ const config = {
             label: "v2.0",
             versions: {
               "1.0": {
-              specPath: "static/vectara-oas.yaml", // path to 1.0 OAS spec
-              outputDir: "versioned_docs/version-1.0/rest-api",
-              label: "1.0",
-              baseUrl: "versioned_docs/version-1.0/rest-api/vectara-oas"
-            }
-           },
+                specPath: "static/vectara-oas.yaml", // path to 1.0 OAS spec
+                outputDir: "versioned_docs/version-1.0/rest-api",
+                label: "1.0",
+                baseUrl: "versioned_docs/version-1.0/rest-api/vectara-oas",
+              },
+            },
           },
         },
       },
@@ -221,28 +221,28 @@ ${content}
           items: [
             {
               label: "Version 2.0 (Latest)",
-              to: "docs/"
+              to: "docs/",
             },
             {
               label: "Version 1.0",
               to: "docs/1.0/",
             },
-          ]
+          ],
         },
         {
           type: "dropdown",
-          label: "API Playground",
+          label: "API Reference",
           position: "left",
           items: [
             {
-              label: "API Playground 2.0 ",
-              to: "docs/rest-api"
+              label: "Version 2.0 (Latest)",
+              to: "docs/rest-api",
             },
             {
-              label: "API Playground 1.0",
+              label: "Version 1.0",
               to: "docs/1.0/rest-api",
             },
-          ]
+          ],
         },
         /* {
           to: "docs/rest-api/",
--- a/www/sidebars.js
+++ b/www/sidebars.js
@@ -103,46 +103,46 @@ module.exports = {
       label: "Security and Data Privacy",
       collapsed: true,
       items: [
-            "learn/data-privacy/privacy-overview",
-            "learn/data-privacy/encryption",
-            {
-              type: "category",
-              label: "Authentication and Authorization",
-              collapsed: true,
-              items: [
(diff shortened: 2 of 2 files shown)


pwoznic:
I'm not sure on renaming Playground to Reference just yet and here's a question and some more context:

- We already have an "API Reference" section in the long-form topics
- Why is the name "API Playground" a problem?
- I ask because this playground term is common in the industry for interactive API documentation, such as with these sites: [DeveloperHub](https://docs.developerhub.io/support-center/try-it-out), [Deepgram](https://playground.deepgram.com/?endpoint=listen&smart_format=true&language=en&model=nova-2), [Adobe](https://developer.adobe.com/experience-cloud/cloud-manager/reference/playground/), dotCMs, hyGraph, Rubrik, etc. Google has something they call an [OAuth Playground](https://developers.google.com/oauthplayground/).
- Google, Meta, Square, Docusign, and ServiceNow use "API Explorer" to (I actually like this name a lot vs Reference and Playground and wanted your thoughts).
- We have many references in the long-form does to the "Playground" as well

cjcenizal:
@pwoznic Great questions! I'll try to explain...

## What's wrong with playground?

Looking at the examples of the other playgrounds, they mostly share a similar user experience in which the user is presented with a "Request input" and a "Response output". The UI emphasizes rapid request/response exploration, similar to [Postman](https://www.postman.com/) but with more guardrails and support in terms of request configuration.

I think the problem is that our current playground doesn't follow this pattern. API v2 introduces three changes that provide an enormous increase in value:

* A well-designed API.
* A comprehensive spec.
* Clear and thoughtful descriptions.

As I integrated API v2 into StreamQueryClient and Console I found myself using the playground as my primary API reference. I didn't issue a single query from it so I didn't use it as a playground at all. I think this is how most folks will use it. As an API reference, it's a real goldmine! 😄 

(Among your examples, the exception is DeveloperHub's playground, but I chalk that up to it not being "productized" -- it's a feature that developers can enable in their docs and I bet they can call it something else, as I'm proposing.)

## Why name it API reference?

I think this is the clearest, simplest description for what this is. If I, as a dev, want to learn about the API's, I want to find a single place to go. I gravitate immediately to "API reference".

For prior art, see [OpenAI docs](https://platform.openai.com/docs/introduction) which distinguish between "playground" (which follows the pattern I described above) and "API reference", and [Cohere](https://docs.cohere.com/) which has just "API reference".

image

I dislike API explorer because it's too ambiguous. I don't know what to expect. It also doesn't resonate with my frame of mind. I'm not in an "exploring" mood. I have a specific problem I'm trying to solve and I just want to look up how a specific bit of the API works so I can get my code working and get through my day.

## What to do about the existing API reference?

I think there's a lot of great content in here. At the same time, I think we should do an exercise where we comb through it and look for opportunities to 1) remove any content that's duplicated elsewhere, and 2) move content into the API playground, Capabilities section, or other sections if it makes more sense elsewhere. I think it will be enlightening to see what's left, which will give us a better idea of exactly what the purpose of this existing API reference section is. I think this will be a fun exercise and I'd love to tag-team it with you if you want.

In the short-term, I think we can disambiguate the names by re-titling this to "API concepts" or something like that.

pwoznic:
> @pwoznic Great questions! I'll try to explain...
> 
> ## What's wrong with playground?
> Looking at the examples of the other playgrounds, they mostly share a similar user experience in which the user is presented with a "Request input" and a "Response output". The UI emphasizes rapid request/response exploration, similar to [Postman](https://www.postman.com/) but with more guardrails and support in terms of request configuration.
> 
> I think the problem is that our current playground doesn't follow this pattern. API v2 introduces three changes that provide an enormous increase in value:
> 
> * A well-designed API.
> * A comprehensive spec.
> * Clear and thoughtful descriptions.
> 
> As I integrated API v2 into StreamQueryClient and Console I found myself using the playground as my primary API reference. I didn't issue a single query from it so I didn't use it as a playground at all. I think this is how most folks will use it. As an API reference, it's a real goldmine! 😄
> 
> (Among your examples, the exception to the pattern is DeveloperHub's playground, but I chalk that up to it not being "productized" -- it's a feature that developers can enable in their docs and I bet they can call it something else, as I'm proposing.)
> 
> ## Why name it API reference?
> I think this is the clearest, simplest description for what this is. If I, as a dev, want to learn about the API's, I want to find a single place to go. I gravitate immediately to "API reference".
> 
> For prior art, see [OpenAI docs](https://platform.openai.com/docs/introduction) which distinguish between "playground" (which follows the pattern I described above) and "API reference", and [Cohere](https://docs.cohere.com/) which has just "API reference".
> 
> image
> 
> I dislike API explorer because it's too ambiguous. I don't know what to expect. It also doesn't resonate with my frame of mind. I'm not in an "exploring" mood. I have a specific problem I'm trying to solve and I just want to look up how a specific bit of the API works so I can get my code working and get through my day.
> 
> ## What to do about the existing API reference?
> I think there's a lot of great content in here. At the same time, I think we should do an exercise where we comb through it and look for opportunities to 1) remove any content that's duplicated elsewhere, and 2) move content into the API playground, Capabilities section, or other sections if it makes more sense elsewhere. I think it will be enlightening to see what's left, which will give us a better idea of exactly what the purpose of this existing API reference section is. I think this will be a fun exercise and I'd love to tag-team it with you if you want.
> 
> In the short-term, I think we can disambiguate the names by re-titling this to "API concepts" or something like that.

Thanks for the detailed explanation and for your perspective as a developer! I see how the playground term may not fully align with how our current users interact with the API, e.g. vs a Postman-like environment/pattern with the rapid request/response interactions.

Let's rename it to API Reference to align with how developer users are probably using it. I also agree about auditing existing content and seeing what we can reorganize and move into this reference to avoid duplication. 

I’m on board with collaborating on the audit and restructuring exercise which will help us identify any gaps, redundancies, and opportunities to enhance our documentation users!

pwoznic (approved):

PR #300, opened 2024-06-28
Fix oauth examples.

nikron:


cjcenizal (approved):
These look right to me but I'm not super familiar with these requests.

PR #312, opened 2024-07-31
Update metadata filter topics

### LINKED ISSUE #311: Metadata filters page doesn't give example of how the metadata is defined
I went to https://docs.vectara.com/docs/learn/metadata-search-filtering/filter-overview to learn more about metadata filters.

That page contains this description:

> Because you can associate this metadata to either the entire document, or to specific parts within it, the scope must be explicitly specified for every metadata reference in the expression. Valid scopes are doc. and part., for document and part-level metadata, respectively.

However, there's no explanation of what these doc and part scopes mean. Can we either link to the pages that contain the relevant information or add examples here?

pwoznic:
Fixes #311 

Updated topics with more examples about document and part scopes and links


www/docs/api-reference/admin-apis/create-corpus.md
```diff
@@ -38,27 +39,39 @@ be changed after corpus creation.
...
 
 In order to reference metadata in [filter expressions](/docs/learn/metadata-search-filtering/filter-overview), the
 referenceable attributes must be declared at creation time in the **filter
 attributes**. This list cannot be changed once the corpus is created.
 
-For information on **custom dimensions**, a Scale-only feature, please see
-[Custom Dimensions](/docs/learn/semantic-search/add-custom-dimensions).
-Like filter attributes, custom dimensions cannot be changed after the corpus
-is created.
+Filter attributes allow you to attach metadata to your data at the document (`doc`)
```
cjcenizal, line 53: Can we address https://github.com/vectara/vectara-docs/issues/311 here? Specifically, I think we need to help the reader understand what `doc` and `part` look like in the context of an actual ingested document's JSON.

cjcenizal (changes requested):

PR #314, opened 2024-08-12
Add User Defined Functions Reranker content

pwoznic:
Updated the Reranking section with information about the new User Defined Functions Rerankers. I also reorganized this section with the other available rerankers.

### DIFF
--- a/www/docs/api-reference/search-apis/reranking.md
+++ b/www/docs/api-reference/search-apis/reranking.md
@@ -1,7 +1,7 @@
 ---
 id: reranking
-title: Reranking
-sidebar_label: Reranking
+title: Rerank Search Results
+sidebar_label: Rerank Search Results
 ---
 
 import Tabs from '@theme/Tabs';
@@ -18,20 +18,23 @@ of query results to achieve a more precise ranking. It employs a machine
 learning model that while slower than the rapid retrieval step, offers more 
 accurate results.
 
-We currently support two rerankers: the [Maximal Marginal Relevance 
-(MMR) Reranker](/docs/api-reference/search-apis/reranking#maximal-marginal-relevance-mmr-reranker) and the new Scale-only [Multilingual Reranker v1](/docs/api-reference/search-apis/reranking#vectara-multilingual-reranker-v1).
+We currently provides the following rerankers: 
+
+* [Multilingual Reranker v1](/docs/learn/vectara-multi-lingual-reranker) `272725719` for accurate results across over 100 languages.
+* [Maximal Marginal Relevance (MMR) Reranker](/docs/learn/mmr-reranker) (`272725718`) for diversifying results while maintaining relevance.
+* [User Defined Function Reranker](/docs/learn/user-defined-function-reranker) `272725722` for custom scoring based on metadata.
 
 ## Enable reranking
 
-To enable reranking, specify the appropriate value for the `rerankerId`. 
+To enable reranking, specify the appropriate value for the `reranker_id`. 
 The MMR reranker ID is `272725718` and the Multilingual Reranker v1 ID is 
 `272725719`. In most scenarios, it makes sense to use the default query `start` 
 value of `0` so that you're reranking all of the best initial results. You can 
-also set `numResults` of the `query` to the total number of documents you wish 
+also set `limit` of the `query` to the total number of documents you wish 
 to rerank. The default value is `25`.
 
-The following example shows the `numResults` and `rerankerId` 
-values in a query. Note that this simplified example intentionally omits  
+The following example shows the `limit` and `reranker_id` 
+values in a query. Note that this simplified example intentionally omits 
 several parameter values.
 
 ```json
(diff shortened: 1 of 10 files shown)


www/docs/api-reference/search-apis/reranking.md
```diff
@@ -18,20 +18,23 @@ of query results to achieve a more precise ranking. It employs a machine
 learning model that while slower than the rapid retrieval step, offers more 
 accurate results.
 
-We currently support two rerankers: the [Maximal Marginal Relevance 
-(MMR) Reranker](/docs/api-reference/search-apis/reranking#maximal-marginal-relevance-mmr-reranker) and the new Scale-only [Multilingual Reranker v1](/docs/api-reference/search-apis/reranking#vectara-multilingual-reranker-v1).
+We currently provides the following rerankers: 
+
+* [Multilingual Reranker v1](/docs/learn/vectara-multi-lingual-reranker) `272725719` for accurate results across over 100 languages.
```
nikron, line 23: Should we call this slingshot? Also we don't really use the number for apiv2. We have "types" of rerankers. Slingshot type is "customer_specific"

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -18,20 +18,23 @@ of query results to achieve a more precise ranking. It employs a machine
...
+* [Maximal Marginal Relevance (MMR) Reranker](/docs/learn/mmr-reranker) (`272725718`) for diversifying results while maintaining relevance.
```
nikron, line 24: this number isn't used in apiv2, it's just type=mmr

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -18,20 +18,23 @@ of query results to achieve a more precise ranking. It employs a machine
...
+* [User Defined Function Reranker](/docs/learn/user-defined-function-reranker) `272725722` for custom scoring based on metadata.
```
nikron, line 25: userfn is the type, number is not used in apiv2

www/docs/api-reference/search-apis/reranking.md
```diff
@@ -18,20 +18,23 @@ of query results to achieve a more precise ranking. It employs a machine
...
 
 ## Enable reranking
 
-To enable reranking, specify the appropriate value for the `rerankerId`. 
+To enable reranking, specify the appropriate value for the `reranker_id`.
```
nikron, line 29: this works differently in apiv2, can you look at the reference and update?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,194 @@
...
+The User Function Language allows you to specify an expression that computes a 
+new score for a search result. There are no statements or variables. The 
+expression definition has access to the search result, and various Vectara 
+provided functions to enable computing new scores for a variety of use cases.
+
+## Types and literals
+
+User Function Language only has double, string, boolean, datetime, and 
+duration as types. You can define double, string, and boolean literals. String 
+literals are enclosed in single quotes (`‘`). To encode a single quote requires
```
nikron, line 34: wrong quote is used here, it should be ' instead of backtick

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,194 @@
...
+   "part_metadata": {
+     "Part level metadata": "metadata"
+   },
+   "document_id": "document id"
+}
+```
+
+#### Get example:
+
+```json
```
nikron, line 108: type is wrong here

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,194 @@
...
+result object. The search result object mirrors the definition of the search 
+result in the HTTP API definition.
+
+
+```json
+{
+  "scores": {
+    "lexical": 234,
+    "neural": .34,
+    "interpolated_score": .34
```
nikron, line 93: this is just interpolated, not interpolated score

nikron: also there is a "last" score which the last reranker or interpolated score computed

nikron (changes requested):

www/docs/api-reference/search-apis/reranking.md
pwoznic: Not sure if Slingshot is just an internal name

www/docs/api-reference/search-apis/reranking.md
nikron: so for customer_specific you need to set the reranker_id

nikron:
Looks good; only error is that customer_specific type does require a reranker id. which would be rnk_{old id}

nikron:
did some edits on the udf-reranker page to fix some examples / flesh out get documentation.

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+     "Part level metadata": "metadata"
+   },
+   "document_id": "document id"
+}
+```
+
+#### Scores object
+
+The scores object allows you to acccess the different scores computed for a 
+search result.
```
eskibars, line 113: Since this section is talking about interpolated and hybrid search scores here, it might warrant a link to the hybrid search docs and/or give a brief primer of what each of these scores actually mean here

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+* `lexical` - The lexical score before the neural score is applied, optimized for 
+  keyword search. This score may be missing. 
+* `neural` - The neural score before the lexical score is applied, providing a 
+  semantic understanding of the query intent.
+* `interpolated` - The computed interpolation between the lexical and neural 
+  scores, blending the keyword and neural. This score may be missing.
+* `previous` - The last score computed by the retrieval pipeline. Most often
```
eskibars, line 120: I think this needs a bit more detail.  I believe this is the score of the previous result, e.g. if the reranker is running on the 6th result, this contains the score of the 5th result post-reranking.  Is that accurate @nikron ?  I imagine folks will be wondering what to do with that score and use cases as well, which we can put an example for

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+  the interpolated score.
+
+#### Get examples
+
+```sql
+get('$.scores.previous') * get('$.part_metadata.boost')
+get('$.document_metadata.reviews[0].score', 0)
+```
+
+#### Lexical score example
```
eskibars, line 130: I think we should craft new examples here because you could essentially do any of the following 3 without any reranker by just changing lambda.  That might be worth explaining in and of itself ("The following is essentially the same as setting lambda = 0 or lambda = 1 ...") and I think we should provide guidance here that _most_ users will likely want to use scores.interpolated and vary lambda in the hybrid search field, but that all scores are available.

But I think the more interesting examples are actually where the score gets combined in some fashion or wholesale replaced to implement metadata sorting.  e.g.:
- get('$.scores.interpolated') + get('$.document_metadata.average_review_score')/10 to bias higher-rated products
- get('$.scores.interpolated') - (to_unix_timestamp(now()) - get('$.document_metadata.date_epochseconds')) to bias toward newness
- if get('$.document_metadata.units_in_stock') > 0 then get('$.scores.interpolated') else 0 to give all items that are out of stock a score of 0 so they show up at the end

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+      "function": "get('$.scores.neural')"
+    }
+  }
+}
+```
+
+#### Interpolated score example
+
+Query example of a hybrid search that balances the scores of neural 
+and lexiccal results that find laptops which are both affordable and suitable
```
eskibars, line 165: lexiccal

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+      "function": "get('$.scores.interpolated')"
+    }
+  }
+}
+```
+
+#### Previous score example
+
+In this example, you want popular coffee shops that have ranked high in 
+previous searches:
```
eskibars, line 183: I don't think this example is accurate?  @nikron

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+perform various operations with datetime values, such as retrieving the 
+current time or converting durations to different units.
+
+| **Function**         | **Description**                                               | **Example**                                 |
+|----------------------|---------------------------------------------------------------|---------------------------------------------|
+| `now()`              | Returns the current time as a datetime.                       | `now()`                                     |
+| `iso_datetime_parse(a)` | Parses an ISO datetime string to a datetime.                 | `iso_date_time_parse('2024-12-04T10:14:50Z')` |
+| `datetime_parse(a,b)`| Parses a datetime string with a format string. The format string follows the Java format string [format](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html). | `datetime_parse('2024 02 09', 'yyyy MM dd')` |
+| `to_unix_timestamp(a)`| Converts a datetime to seconds from the epoch.               | `to_unix_timestamp(now())`         |
+| `seconds(a)`         | Converts a duration to the number of seconds.                 | `seconds(seconds(50))`             |
```
eskibars, line 210: @nikron @pwoznic is this double-casting accurate?  e.g. seconds(seconds()), minutes(minutes()) and hours(hours()) or should these be singular?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,285 @@
...
+      "reranker": {
+        "type": "userfn",
+        "function": "get('$.scores.previous') * get('$.document_metadata.boost')"
+      }
+   }
+}
+```
+## gRPC example
+
+This example of a toy shows how ranking in an ecommerce system would work:
```
eskibars, line 269: A toy?

eskibars (changes requested):
Thanks for the draft @pwoznic.  I left some comments/suggestions

www/docs/learn/user-defined-function-reranker.md
nikron: no, it's the last previous reranker score if there are two rerankers scored together, or the final score of the chain. his name was too confusing and i'm eliminating it

www/docs/learn/user-defined-function-reranker.md
nikron: removing previous

www/docs/learn/user-defined-function-reranker.md
nikron: Yes this is accurate; same function can cast a number to a double and vice versa.

nikron:
@pwoznic we decided to remove the scores object. i added more docs on scores

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -0,0 +1,216 @@
+---
+id: user-defined-function-reranker
+title: User Defined Function Reranker
+sidebar_label: User Defined Function Reranker
+---
+
+import {Config} from '@site/docs/definitions.md';
+
+The User Defined Functions Reranker enables users to define custom reranking
```
cjcenizal, line 9: Let's preface every doc with a simple, accessible description of the problem that's being solved with the feature described by the doc. For example:

```
Our out-of-the-box rerankers are effective for general use cases, but some specific use cases require fine-grained control over how search results are ordered. For example, bubbling recently-added documents to the top, or limiting search results to a specific geolocation.
```

PR #323, opened 2024-08-22
Add more UDF examples

### LINKED ISSUE #321: Provide example of how to sort using the UDF reranker
Per @eskibars we can use the UDF reranker to sort search results based on filter metadata. Can we update https://docs.vectara.com/docs/learn/user-defined-function-reranker with an example of how to do this?

pwoznic:
Fixes #321 

Adds a few more examples about the flexibility of UDF

### DIFF
--- a/www/docs/learn/user-defined-function-reranker.md
+++ b/www/docs/learn/user-defined-function-reranker.md
@@ -4,6 +4,8 @@ title: User Defined Function Reranker
 sidebar_label: User Defined Function Reranker
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 import {Config} from '@site/docs/definitions.md';
 
 Our out-of-the-box rerankers are effective for general use cases, but some 
@@ -15,15 +17,21 @@ The User Defined Functions Reranker enables users to define custom reranking
 functions using document-level metadata, part-level metadata, or scores 
 generated from the request-level metadata. To use this reranker, set the 
 `type` to `userfn` in a query and specify a string within the 
-`reranker` section of the query. This string syntax is custom and similar to our
+`reranker` section of the query. This string syntax is custom and similar to our 
 SQL-like [filter expressions](/docs/learn/metadata-search-filtering/filter-overview).
 
-This new user-defined functions reranker allows for a wide range of use cases:
+With the flexibility to modify scores based on metadata, conditions, and 
+custom logic, enterprises can craft highly tailored search experiences that 
+meet specific business needs. This new User Defined Functions reranker allows 
+for a wide range of use cases:
 
 * **Recency bias**: Prioritize the most recent results in cases where answers 
-  based on older data are less relevant than newer data.
-* **Location bias**: Prioritize results closer to the location of the user.
-* **E-commerce bias**: Prioritize promotional and sponsored merchandise.
+  based on older data are less relevant than newer data. Examples include news 
+  and current events searches, stock market queries, and recruitment searches. 
+* **Location bias**: Prioritize results closer to the location of the user such 
+  as local business searches, real estate listings, and event queries.
+* **E-commerce bias**: Prioritize promotional and sponsored merchandise for 
+  sale promotions and new product launches.
 
 ## Language definition
 
@@ -43,6 +51,7 @@ literals are enclosed in single quotes (`'`). To encode a single quote requires
 two repeated single quotes (`''`).
 
 ### Duration
+
 Datetime operations such as subtraction result in `duration`s instead of integers
 unlike many SQL languages. This allows precise handling of datetime operations
 when doing recency boosting or other time based modifications of the score.
@@ -75,7 +84,7 @@ Operator precedence is in the previous listed order.
 (1 + 2 + 3) / 6
 ```
 
-## If Expression
+## If expression
 
 Expressions may also be a conditional if expression. Both results of the 
 condition must be defined. The grammar for the if expression is 
@@ -93,7 +102,7 @@ Functions are in the normal `function(argument1, argument2)` syntax. All
 functions are predefined by Vectara, and retrieving properties of the current 
 search result is done with the function `get.`
 
-### Get Function
+### Get function
 
 The `get` function allows you to retrieve properties of the result. The `get` 
 argument is a [JSONPath](https://en.wikipedia.org/wiki/JSONPath) string that retrieves values 
@@ -101,7 +110,7 @@ from a search result object. JSON paths that do not exist return null.
 The `get` function can also optionally take a second argument as a default 
 value if the JSONPath is null.
 
-The search result object is similiar to the definition of the search 
+The search result object is similar to the definition of the search 
 result in the HTTP API definition. What follows is the schema for the search 
 result object. 
 
@@ -152,7 +161,7 @@ The `seconds`, `minutes`, and `hours` functions allow you to convert a number to
 
 ## Math functions
 
-The following table list the math functions available that allow you to perform 
+The following table lists the math functions available that allow you to perform 
 various mathematical operations, such as calculating absolute values, powers, 
 and logarithms.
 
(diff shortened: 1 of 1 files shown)


www/docs/learn/user-defined-function-reranker.md
```diff
@@ -184,7 +184,35 @@ and logarithms.
 get('$.score') * 1 / as_days(iso_datetime_parse(get('$.document_metadata.publication_date')) - now())
 ```
 
-## API v2 example
+## gRPC example
+
+This example shows how ranking in an ecommerce system would work:
+
+
+```grpc
```
cjcenizal, line 192: I don't think we need this example to be in grpc or any specific language. Let's just show the function by itself. Let's also show an example of a document with the metadata that's being used, and let's annotate it to explain the various metadata.

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -184,7 +184,35 @@ and logarithms.
...
+ request = serving_pb2.QueryRequest(
+        query=query, num_results=fetch)
+ request.reranking_config.reranker_id = 272725722
+ request.reranking_config.user_function = \
+     "get('$.score') + log10(get('$.document_metadata.publish_ts')) + log(get('$.document_metadata.customer_review_stars')) + get('$.document_metadata.promoted')"
+```
+
+This example modifies the score with the following options:
+
+* Recency bias: Adds the log of the publication timestamp (`get('$.document_metadata.publish_ts`))
```
cjcenizal, line 202: Can we explain the why behind each of these steps?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -184,7 +184,35 @@ and logarithms.
...
+* Popularity bias: Adds the log of the customer review, which can be 1 to 5 
+  stars (`get('$.document_metadata.customer_review_stars')`)
+* Sponsored promotions bias: Adds the value of the promoted variable (`get('$.document_metadata.promoted')`)
+
+## API v2 examples
```
cjcenizal, line 207: Why is this title needed?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -200,22 +228,79 @@ metadata:
...
+
+`if get('$.document_metadata.units_in_stock') > 0 then get('$.score') else -999999`
+
+
+### Boost products that have a higher score
+
+Boost products that have a higher review score, where `average_review_score` is 
+a metadata field with values between `0` and `10`:
+
+`get('$.score') + get('$.document_metadata.average_review_score') / 10`
```
cjcenizal, line 257: Why is it important to normalize this value by dividing by 10?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -200,22 +228,79 @@ metadata:
...
+
+
+### Boost content by type
+
+Boost sections in product documentation that contain `Technical Specifications` 
+by 1.6, ensuring that these specifications are:
```
cjcenizal, line 263: "Ensuring that these specifications are" -- I think we're missing something from the end of this sentence.

Also, why multiply by 1.6?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -200,22 +228,79 @@ metadata:
...
 
-* Recency bias: Adds the log of the publication timestamp (`get('$.document_metadata.publish_ts`))
-* Popularity bias: Adds the log of the customer review, which can be 1 to 5 
-  stars (`get('$.document_metadata.customer_review_stars')`)
-* Sponsored promotions bias: Adds the value of the promoted variable (`get('$.document_metadata.promoted')`)
+In multilingual documents, boost sections written in French by 1.5, ensuring 
+that users see content in this preferred language when searching for 
+`tourism industry reports`.
+
+`if(get('$.part_metadata.lang') == 'fra', get('$.score') * 1.5, get('$.score'))`
```
cjcenizal, line 283: Why multiply by 1.5?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -200,22 +228,79 @@ metadata:
...
+`1 / (now() - iso_datetime_parse(get('$.document_metadata.publication_date')))`
+
+### Prioritize low-rated documents
+
+Sometimes you **do not** want to prioritize high-rated documents. Instead, 
+you want to know more about bad reviews to identify issues with your products. 
+For example, boost `customer support feedback` documents with bad reviews by `1.6` 
+(`score < 3`) to help identify problem areas based on negative feedback.
+
+`if(get('$.document_metadata.customer_review_score') < 3, get('$.score') * 1.5, get('$.score'))`
```
cjcenizal, line 301: Why multiply by 1.5?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -200,22 +228,79 @@ metadata:
...
+
+These are just a few examples of how the User Defined Functions (UDF) reranker
```
cjcenizal, line 303: Since this is docs, not a blog post, it seems odd and somewhat buried for this content to be placed at the end. It seems unnecessary so I think we can cut it but if you think it's important then I suggest placing it in the introduction, at the top of the page.

cjcenizal (commented):
Thanks for adding this information @pwoznic ! The examples that I understood were really helpful. There were a few places where I was confused though.

www/docs/learn/user-defined-function-reranker.md
eskibars: This was an example I provided.  Users frequently introduce some arbitrary multiplications and divisions in these types of reranking functions.  Divide by some reasonably large number (5-50) is reasonably common in these scenarios when you have a score usually in the range of 0-5 (which isn't guaranteed to be the case with BM25, which has unbounded upper maximum scores, but is "normal" for some data sets/queries.  I'm not sure we can really prescribe a good mechanism for picking these, because it depends on the types of queries, the type of data, and "how much" a user wants to bias.

We'll be able to make a better recommendation on this page around the effects of these additions/multipliers once we have score normalization on the platform because the first 2 of the "depends" can largely be normalized, but until then it's going to be completely arbitrary

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -177,45 +186,217 @@ and logarithms.
...
 ```json
 {
-   "query": "query",
-   "search": {
-      "reranker": {
-        "type": "userfn",
-        "function": "get('$.score') * get('$.document_metadata.boost')"
-      }
-   }
+  "id": "DD-2025-ELECTRONICS-FALL",
```
eskibars, line 203: It's possible that this could be a single document, but the more common document modeling would leave each of these 3 products as separate documents.  I'm OK with pushing this forward to get something out now, but I think we should reformat this set of examples quickly to be focused on documents instead of document parts

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -177,45 +186,217 @@ and logarithms.
...
+  such as top-tier products.
+
+### Boost on metadata
+
+This example multiplies the base score by a boost value from the document 
+metadata:
+
+`get('$.score') * get('$.document_metadata.boost')`
+
+### Boost products that have a higher score
```
eskibars, line 342: "Score" is ambiguous here: should we rephrase to "customer rating" here and in the following paragraphs?

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -177,45 +186,217 @@ and logarithms.
...
+higher customer review scores by factoring the review into the overall 
+relevance score. This approach is useful in e-commerce, service platforms, or 
+content recommendation systems where user feedback plays an important role in 
+ranking and relevance.
+
+Many customer review systems use a scale from 0 to 10. In this example, 
+dividing by 10 normalizes the impact of customer reviews on the final score and 
+prevents overweighting the review score. If the scale was 0 to 5, you would 
+divide by 5. This way the customer feedback acts as a boost rather than 
+dominating the search rankin:
```
eskibars, line 355: s/rankin/ranking/

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -177,45 +186,217 @@ and logarithms.
...
+
+`get('$.score') + get('$.part_metadata.customer_review_stars', 0) / 10`
+
+### Boost content by type
+
+Boost sections by for queries about `product manual for model X` in product
```
eskibars, line 361: Extraneous `by`

www/docs/learn/user-defined-function-reranker.md
```diff
@@ -177,45 +186,217 @@ and logarithms.
...
+prevents overweighting the review score. If the scale was 0 to 5, you would 
+divide by 5. This way the customer feedback acts as a boost rather than 
+dominating the search rankin:
+
+`get('$.score') + get('$.part_metadata.customer_review_stars', 0) / 10`
+
+### Boost content by type
+
+Boost sections by for queries about `product manual for model X` in product 
+documentation that contain `Technical Specifications`. This example uses `1.5`
```
eskibars, line 362: s/contain/contains/

eskibars (approved):
Thanks @pwoznic .  I left some comments here; feel free to take what you feel is best

PR #327, opened 2024-09-18
feat: setup fern to use latest OpenAPI

dsinghvi:
This PR updates your Fern Config to be inside of your docs repository so that Fern can leverage your latest OpenAPI spec.

### DIFF
--- a/.github/workflows/ts-sdk.yml
+++ b/.github/workflows/ts-sdk.yml
@@ -1,0 +1,29 @@
+name: Release TypeScript SDK
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "The version of the SDKs that you would like to release"
+        required: true
+        type: string
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/ts-sdk.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Download Fern
+        run: npm install -g fern-api
+
+      - name: Release SDKs
+        env:
+          FERN_TOKEN: ${{ secrets.FERN_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          fern generate --group ts-sdk --version ${{ inputs.version }} --log-level debug
```
nikron, line 29: do we need debug level logs?

nikron (approved):

.github/workflows/ts-sdk.yml
dsinghvi: Let's keep them till you become very comfortable with Fern, if thats okay with you!

PR #330, opened 2024-09-24
Update Chain Reranker topic about the array and examples

pwoznic:
Updated the description of the chain reranker and also added another example

### DIFF
--- a/www/docs/learn/chain-reranker.md
+++ b/www/docs/learn/chain-reranker.md
@@ -105,7 +105,7 @@ products.
     "rerankers": [
       {
         "type": "customer_specific",
-        "reranker_id": "rnk_272725719"
+        "reranker_name": "Rerank_Multilingual_v1"
       },
       {
         "type": "mmr",
--- a/www/docs/learn/chain-reranker.md
+++ b/www/docs/learn/chain-reranker.md
@@ -9,12 +9,13 @@ import {vars} from '@site/static/variables.json';
 
 When refining query results where relevance and diversity are critical to 
 getting the best answers, relying on a single ranking strategy may not be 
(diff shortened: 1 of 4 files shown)


www/docs/learn/chain-reranker.md
```diff
@@ -9,18 +9,47 @@ import {vars} from '@site/static/variables.json';
...
+To enable the chain reranker, set the `reranker` object `type` to `chain` and then 
+specify the multiple rerankers that you want to combine in a `rerankers` array. 
+You can specify up to 50 rerankers in a chain.
+
+## Chain reranker examples
+
+In this example, the Vectara Multilingual Reranker first refines the results 
+across over 100 languages, while the [User Defined Function reranker](/docs/learn/user-defined-function-reranker) 
+then boosts relevance based on metadata for popularity. This simple chain 
+example can be useful for multilingual content platforms such as blogs, news
```
eskibars, line 50: The examples here should be refined for the same reasons as above with the multilingual reranker.  It doesn't really have to do with multilingual content -- it just gives better ranking order than Boomerang (and it just so happens that it can do that across 100 languages)

eskibars (changes requested):
Mostly minor comments, but 1 that I think we should review: we need to revise how we're talking about the multilingual reranker.  I've left comments in here to give some additional background for you, but we should probably discuss further to make sure you're up to speed on this feature.  Can you also double check the rest of the docs to see if we mention this elsewhere?

pwoznic:
I added the cutoffs property and updated the multilingual reranker verbiage in the latest commit.

eskibars (approved):

PR #345, opened 2024-10-16
Update docs for handling filenames with non-ASCII characters in API v2

pwoznic:
Added a note on URL encoding filenames with non-ASCII characters when uploading files via API v2 to follow web standards.


david-vectara:
Change looks good, thanks Paul.

david-vectara (approved):

PR #349, opened 2024-10-24
fix: Incorrect syntax in the null score usage chain example

pwoznic:
Added "document_" to metadata in the chain example


ofermend (approved):
LGTM

PR #358, opened 2024-11-22
fix(bug): Update how metadata structuring is handled in API v2

### LINKED ISSUE #357: Metadata structuring is not handled in the same way as the docs for API v2
### Pages

https://docs.vectara.com/docs/api-reference/search-apis/interpreting-responses/metadata

### Description

The way in which metadata responses are created ensures that pure numbers will be returned as numbers, the same thing with booleans and JSON.

Here's the issue:
```
      "part_metadata": {
        "speaker": "Deep Thought",
        "lang": "eng",
        "section": "2",
        "offset": "316"
      },
      "document_metadata": {
        "author": "Douglas Adams",
        "publicationyear": "1979"
      },
``` 

will actually be

```
      "part_metadata": {
        "speaker": "Deep Thought",
        "lang": "eng",
        "section": 2,
        "offset": 316
      },
      "document_metadata": {
        "author": "Douglas Adams",
        "publicationyear": 1979
      },
``` 

It looks like further down in the same document this is addressed correctly. 

I would also recommend adding this as part of https://docs.vectara.com/docs/migration-guide-api-v2 as it was handled differently in v1.

pwoznic:
Fixes #357 

Updated Reading Metadata and the v1 to v2 migration guide

### DIFF
--- a/www/docs/api-reference/search-apis/interpreting-responses/metadata.md
+++ b/www/docs/api-reference/search-apis/interpreting-responses/metadata.md
@@ -13,10 +13,9 @@ document has a `type` parameter that determines the format of the document
 as `core` or `structured`. The `core` type has `document_parts` and the `structured` 
 type has `sections`. Both can be nested and both can contain separate `metadata`, 
 including some metadata that <Config v="names.product"/> will auto-generate. 
-A good example of this is that you could have a document which has some global 
-attributes like the `URL` or `owner` but individual sections will have a `section` 
-attribute and a `lang`.
 
+For example, a document might have global attributes such as the `URL` or `owner` 
+but individual sections have a `section` attribute and a `lang`.
 
 Here's an example response with different metadata at these different levels:
 
@@ -29,12 +28,12 @@ Here's an example response with different metadata at these different levels:
       "part_metadata": {
         "speaker": "Deep Thought",
         "lang": "eng",
-        "section": "2",
-        "offset": "316"
+        "section": 2,
+        "offset": 316
       },
       "document_metadata": {
         "author": "Douglas Adams",
-        "publicationyear": "1979"
+        "publicationyear": 1979
       },
       "document_id": "hitchhikers-guide",
       "request_corpora_index": 0
@@ -44,8 +43,8 @@ Here's an example response with different metadata at these different levels:
       "score": 0.13511724770069122,
       "part_metadata": {
         "lang": "eng",
-        "section": "17",
-        "offset": "171"
+        "section": 17,
+        "offset": 171
       },
       "document_metadata": {
         "author": "Dr. Seuss"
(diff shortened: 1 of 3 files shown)


www/docs/api-reference/search-apis/interpreting-responses/metadata.md
```diff
@@ -64,29 +63,46 @@ metadata. The reason for this split is that there may be multiple sections
 from the same document in the response, and this allows for deduplication of
 document-level metadata, which can reduce the total time for the response.
 
+## Metadata type consistency
+
+The metadata type conversion applies only to the query responses. Metadata 
+remains unconverted during the document upload process, even when using API v2: 
+
+* **Numbers** are returned as numbers (for example, `section: 2`, `publicationyear: 1979`).
```
abhishekpradhan, line 71: Might be a good idea to give an idea of what converted here.

Anything that follows the regex: `-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?`

| Input       | Matches? | Explanation                                   |
|-------------|----------|-----------------------------------------------|
| `123`       | ✅        | Valid integer.                               |
| `0`         | ✅        | Valid zero.                                  |
| `-456`      | ✅        | Valid negative integer.                      |
| `3.14`      | ✅        | Valid decimal number.                        |
| `-0.001`    | ✅        | Valid negative decimal.                      |
| `2e10`      | ✅        | Valid scientific notation.                   |
| `-1.23E-4`  | ✅        | Valid negative number in scientific notation.|
| `.5`        | ❌        | Invalid (missing leading integer).           |
| `1e`        | ❌        | Invalid (missing exponent value).            |
| `1.2.3`     | ❌        | Invalid (multiple decimal points).           |
| `-`         | ❌        | Invalid (missing digits).                    |

www/docs/api-reference/search-apis/interpreting-responses/metadata.md
```diff
@@ -64,29 +63,46 @@ metadata. The reason for this split is that there may be multiple sections
...
+* **Booleans** are returned as booleans.
```
abhishekpradhan, line 72: This only works for exactly `true` or `false`. Regex: `^(true|false)$` Added some examples here: https://github.com/vectara/vectara-docs/issues/357#issuecomment-2494776275

www/docs/api-reference/search-apis/interpreting-responses/metadata.md
```diff
@@ -64,29 +63,46 @@ metadata. The reason for this split is that there may be multiple sections
...
+* **JSON objects** maintain their native structure.
```
abhishekpradhan, line 73: Added regex and examples for this one too: https://github.com/vectara/vectara-docs/issues/357#issuecomment-2494785819

abhishekpradhan (changes requested):
I think it's a good idea to add additional information, otherwise it's still confusing. Documenting the regex and examples will allow users of this to easily account for this. 

I had a follow question asking what exactly it converted.

www/docs/api-reference/search-apis/interpreting-responses/metadata.md
abhishekpradhan: JSON numbers, not sure if it's useful to include this here or not: https://json-schema.org/understanding-json-schema/reference/numeric#number

abhishekpradhan (approved):

PR #362, opened 2024-12-02
feat(new) Add query observability feature to docs

pwoznic:
Added documentation for the new query observability feature. This contains the long-form updates and the API reference updates will be next.


www/docs/api-reference/query-history-apis/get-query-histories.md
```diff
@@ -0,0 +1,24 @@
...
+sidebar_title: List Query Histories API Definition
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+import {vars} from '@site/static/variables.json';
+import {Config} from '@site/docs/definitions.md';
+
+The List Query Histories API allows you to retrieve, update, and manage query
```
omnomagonz, line 13:
```suggestion
The List Query Histories API allows you to retrieve, update, and manage query 
```

Are we firm on this name? It sounds a bit odd to me. Could we consider: Query history API or Query history list API? 

Non-blocking suggestion 😄

www/docs/api-reference/query-history-apis/get-query-histories.md
```diff
@@ -0,0 +1,24 @@
...
+import CodeBlock from '@theme/CodeBlock';
+import {vars} from '@site/static/variables.json';
+import {Config} from '@site/docs/definitions.md';
+
+The List Query Histories API allows you to retrieve, update, and manage query 
+history for a specific corpus. This API is particularly useful for tracking 
+query performance, debugging individual queries, and retrieving detailed 
+information such as the call stack of a query execution.
+
+## List Query Histories Request and Response
```
omnomagonz, line 18: This may just be a me thing today, but my brain doesn't grok this naming well.

To me even changing it to "List Query History" might resonate more than the plural "Histories". It's a nit, so I'm not firm here

www/docs/api-reference/query-history-apis/get-query-histories.md
```diff
@@ -0,0 +1,24 @@
...
+
+To get a history of a specific query, send a GET request to 
+`/v2/queries`. You can specify the `corpus_key`, `chat_id`, and the 
+`limit` which is the maximum number of historical queries to list. The 
+response includes an array of query histories.
```
omnomagonz, line 23:
```suggestion
response includes an array of previous queries.
```

Non-blocking

www/docs/api-reference/query-history-apis/get-query-history.md
```diff
@@ -0,0 +1,29 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+import {vars} from '@site/static/variables.json';
+import {Config} from '@site/docs/definitions.md';
+
+The Get Query History API allows you to retrieve detailed history about a 
+specific query that was made against a corpus. The response includes detailed
```
omnomagonz, line 14:
```suggestion
The Get Query History API allows you to retrieve detailed history about a 
specific query that was made against a corpus. The response includes detailed 
```

This may be naming convention, so this is another non-blocking suggestion: could we consider changing this to be (imo) more distinct vs the other `List Query Histories` API? An alternative here could be: Query Analysis API or Single Query History API.

I think something that can help further differentiate between this single query history object and the other list of multiple queries object could be helpful

www/docs/console-ui/configure_queries.md
```diff
@@ -0,0 +1,112 @@
...
+query experience to your specific needs:
+
+### Save query to history
+
+When experimenting with configurations and running queries against a corpus, 
+Vectara lets you log your queries into a query history. This is important for 
+troubleshooting issues, inspecting past queries, and optimizing configurations. 
+By surfacing data like query latency, search results, reranking, and 
+generative response times, users can better understand how our system performs 
+relative to their business goals.
```
omnomagonz, line 53:
```suggestion
generative response times, users can better understand how to fine-tune Vectara to meet their business goals.
```

non-blocking

www/docs/console-ui/configure_queries.md
```diff
@@ -0,0 +1,112 @@
...
+
+### View the history of query
+
+1. Experiment with different query configurations by asking questions of 
+   your data with **Send query**.
```
omnomagonz, line 58:
```suggestion
1. Ask questions of your data with **Send query**.
```

Non-blocking.

This reads a bit odd to me. I think the intent here is to say: "First, submit a query", but it sort of jumps ahead with "experiment with different queries".

My mind is interpreting this as out of order. I think if we lead with "first, submit a query" and then maybe add a separate step of "use this information to optimize your configuration and submit a new query" this could help

omnomagonz (approved):
@pwoznic some non-blocking suggestions. Overall the docs look great and most (all?) of my comments are very subjective 😄

www/docs/api-reference/query-history-apis/get-query-histories.md
pwoznic: Not sure if firm (we can run by team for sure), but FWIW I'm going by the yaml descriptions where one is for a specific query and the other is to list a group of multiple queries.

image

mrderyk: FWIW CJ updated the console copy to say “query histories”

www/docs/api-reference/query-history-apis/get-query-histories.md
mrderyk: Same comment here: if consistency with the UI makes any difference, “Query Histories” is what we use in console.

what you said makes a lot of sense though, @omnomagonz

www/docs/console-ui/inspect_queries.md
```diff
@@ -0,0 +1,53 @@
...
+* **Search:** Shows the initial retrieval of relevant data from the corpus.
+* **Reranking:** Refines the order of retrieved results for improved relevance.
+* **Generation:** Produces the final output, including citations and a summary from the LLM.
+* **Factual consistency scoring:** Evaluates the factual accuracy of the generated output against 
+  retrieved data.
+
+Click **Expand all** to reveal detailed information for each execution stage, 
+including specific parameters.
+
+Select **Load into Query tab** to go back to query configuration so that you
```
mrderyk, line 52: This is a little unclear to me. How about something like “…to load this configuration into your query tab”?

It just reiterates the button text, but I think it may be clearer than saying “go back to query configuration”.

www/docs/api-reference/query-history-apis/get-query-histories.md
omnomagonz: I think that naming convention makes sense with a UI where there's more (visual) context, so this may be a non-issue if our users are familiar with the console/broader context of this.

I think if there's a standalone experience where a user comes across this API it may have some additional cognitive load to understand what exactly it means or does 🙂

PR #364, opened 2024-12-04
docs(update): Update User-Defined Function reranker docs with new all_scores and index functions

pwoznic:
Added the get('$.all_scores') and get('$.index') functions to the UDF section


husseinVectara (approved):
looks good to me

husseinVectara (commented):
this looks right, i need to verify

PR #378, opened 2025-01-03
feat(new): Add documentation for the new Knee Reranking feature

pwoznic:
Added and updated topics for the new Knee Reranker feature:

- Added Knee Reranker to the Reranking section
- Updated the Reranking parent topic
- Updated the Query API long-form topic
- Updated the Release Notes

### DIFF
--- a/www/docs/learn/knee-reranker.md
+++ b/www/docs/learn/knee-reranker.md
@@ -1,0 +1,77 @@
+---
+id: knee-reranker
+title: Knee Reranker
+sidebar_label: Knee Reranker
+---
+
+import {Config} from '@site/docs/definitions.md';
+
+Retrieval Augmented Generation (RAG) systems often struggle with determining 
+optimal cutoff points for query results. Fixed score thresholds provide a 
+simple solution but lack the flexibility to adapt to varying score 
+distributions across queries. The knee reranker addresses this challenge by 
+detecting natural boundaries between relevant and irrelevant results 
+automatically.
+
+The knee reranker combines statistical analysis with configurable parameters 
+to provide intelligent, adaptive filtering. Designed specifically to work 
+after the Slingshot reranker, it analyzes score patterns to identify 
+significant drops in relevance while maintaining safeguards against 
+over-aggressive filtering. For more details about how this reranker works, see 
+this blog post.
+
+## Enable the knee reranker
+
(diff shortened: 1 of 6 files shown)


www/docs/learn/knee-reranker.md
```diff
@@ -0,0 +1,77 @@
...
+import {Config} from '@site/docs/definitions.md';
+
+Retrieval Augmented Generation (RAG) systems often struggle with determining 
+optimal cutoff points for query results. Fixed score thresholds provide a 
+simple solution but lack the flexibility to adapt to varying score 
+distributions across queries. The knee reranker addresses this challenge by 
+detecting natural boundaries between relevant and irrelevant results 
+automatically.
+
+The knee reranker combines statistical analysis with configurable parameters
```
husseinVectara, line 16: hey paul, could yo please rephrase the 'reranker' terminology to reranking instead? Its less confusing to users i would say. Please see shane's updated blogpost for reference!

www/docs/learn/knee-reranker.md
```diff
@@ -0,0 +1,77 @@
...
+* **Balanced defaults:** General-purpose filtering where precision and recall 
+  are equally important, such as content recommendation systems or general 
+  RAG workflows.
+  
+  `"user_function": "knee(0.5, 0.2)"`
+* **High precision:** Applications requiring strict relevance, where irrelevant 
+  results must be aggressively filtered out
+  
+  `"user_function": "knee(0.8, 0.5)"`
+* **Broader exploration:** Exploratory research or academic queries, where
```
husseinVectara, line 73: I wouldn't steer awya for giving exact examples of these parameters, maybe lets keep this to general descriptions of high sensitivity, means small drops can trigger knee, and high early bias places stronger emphasis on earlier knees if they are similar in magnitude

husseinVectara (changes requested):
Hey paul, I think this is good, but a little rewording to not specifically say knee reranker, but rather call it knee reranking (as its a functions, not a reranker in and of itself is necessary. another note was just the examples you gave, I'm not sure those examples are exactly accurate so id say maybe lets change those, or we can think of exact examples if you really want to have examples in the doc

pwoznic:
Made several changes and also fixed some links that prevented the preview.

PR #383, opened 2025-01-10
docs(ia): Add a dedicated changelog for documentation updates, API enhancements and bug fixes

pwoznic:
Added a CHANGELOG to the documentation site.

### DIFF
--- a/www/docs/DocCHANGELOG.md
+++ b/www/docs/DocCHANGELOG.md
@@ -11,6 +11,38 @@ import { Spacer } from "@site/ui/Spacer";
 import { TopicButton } from "@site/ui/TopicButton";
 
 
+---
+
+# January 2025
+
+### ✨ New Features
+- Update HHEM documentation for HHME-2.2 (#379)
+- Add documentation for the new Knee Reranking feature (#378)
+- Add new HHEM supported languages and some content enhancements (#352)
+
+### 📚 API Changes
+- No changes
+
+### 📝 Documentation or Feature Enhancements
+- Add HHEM 2.2 enhancements to Release Notes (#381)
+- Add new HHEM supported languages and some content enhancements (#352)
+
+### 📐 Information Architecture Changes
+- No changes
+
+### 🐛 Documentation Fixes
+- No changes
+
+### 🗑 Deprecated
+- No changes
+
+### 📋 Example Updates
+- No changes
+
+### 🛠 Documentation Platform Updates
+- No changes
+
+
 
 ---
 
@@ -477,3 +509,4 @@ import { TopicButton } from "@site/ui/TopicButton";
 
 ### 🛠 Documentation Platform Updates
 - No changes
+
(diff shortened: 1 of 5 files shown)


www/docs/DocCHANGELOG.md
```diff
@@ -0,0 +1,516 @@
...
+import { Grid } from "@site/ui/Grid";
+import { Spacer } from "@site/ui/Spacer";
+import { TopicButton } from "@site/ui/TopicButton";
+
+
+---
+
+# January 2025
+
+### ✨ New Features
```
omnomagonz, line 18: Non-blocking: We might re-consider the use of emojis. This is a highly subjective opinion, but they distract me particularly if a section has little-to-no content. It also seems particularly distracting and repetitive on the right side panel.

Food for thought! I like the intent of making it more playful or engaging so I may just not be the audience for this 😄

www/docs/DocCHANGELOG.md
```diff
@@ -0,0 +1,516 @@
...
+- Add new HHEM supported languages and some content enhancements (#352)
+
+### 📚 API Changes
+- No changes
+
+### 📝 Documentation or Feature Enhancements
+- Add HHEM 2.2 enhancements to Release Notes (#381)
+- Add new HHEM supported languages and some content enhancements (#352)
+
+### 📐 Information Architecture Changes
```
omnomagonz, line 30: Non-blocking: These seem a bit duplicative to me. What if we did: Feature enhancements and Documentation changes as two separate categories?

In my opinion: "New Features" and "Feature Enhancements" are very similar. So we could even consider something like "Product changes" or something to encompass both?

Similarly, "Documentation" and "Information Architecture Changes" seem similar. Maybe just "Documentation" or "Documentation Changes"? This assumes Info. Arch is related to documentation, but if it's something else then we can ignore this 👍

www/docs/DocCHANGELOG.md
```diff
@@ -0,0 +1,516 @@
...
+- No changes
+
+### 🐛 Documentation Fixes
```
omnomagonz, line 33: Non-blocking: Same comment as above. This could maybe fit under a singular category like "Documentation changes" and then you could do sub-classifications, e.g.:

**Documentation changes**
- Added: new page
- Added: other page
- Changed: section title
- Changed: page location
- Changed: python example for single query API endpoint
- Fix: something
- Fix: something else

www/docs/DocCHANGELOG.md
```diff
@@ -0,0 +1,516 @@
...
+- No changes
+
+### 🗑 Deprecated
+- No changes
+
+### 📋 Example Updates
```
omnomagonz, line 39: Non-blocking: same comment as the previous two. Could this be consolidated?

www/docs/DocCHANGELOG.md
```diff
@@ -0,0 +1,516 @@
...
+- No changes
+
+### 🛠 Documentation Platform Updates
```
omnomagonz, line 42: Non-blocking: Same as previous comments. Could this be consolidated? I'm also generally unclear what specifically this refers to

omnomagonz (approved):

PR #394, opened 2025-01-22
deprecate: Uploading JSON in API v2

pwoznic:
Updated the migration guide


www/docs/api-reference/indexing-apis/file-upload/supported_file_formats.md
```diff
@@ -28,15 +28,13 @@ types are supported:
 
 ## Semi-structured documents
 
-In addition, the [upload endpoint](/docs/api-reference/indexing-apis/file-upload/file-upload) supports
-sending semi-structured documents through this endpoint that reflect a
-`Document` proto message.  Those can be sent in the following formats:
+In API v1, the upload endpoint supports sending semi-structured documents through
```
cjcenizal, line 31: Generally I think we should limit our references to prior versions to the context of migrating or otherwise dealing with a breaking change. It's not clear to me if that's the context in this case. What is the context of this reference to v1?

www/docs/api-reference/search-apis/interpreting-responses/highlighting.md
```diff
@@ -91,9 +91,8 @@ want to include content from other sections.
 
 If you created the sections yourself, you may choose to just replay the
 sectioning logic at query time and use as much of the additional sections as
-desired.  However, if you aren't certain as to the section numbers and/or if
-you uploaded documents using the
-[file upload API](/docs/api-reference/indexing-apis/file-upload/format-for-upload), then you might need
+desired. However, if you aren't certain as to the section numbers and/or if
```
cjcenizal, line 94: I think it makes sense to treat API names as proper nouns to make them easier to identify, e.g. "File Upload API". WDYT?

www/docs/migration-guide-api-v2.md
```diff
@@ -171,6 +171,16 @@ changes carefully and the action items to use the V2 endpoints.
 **Action items:** Update your indexing requests to use the new unified endpoint, 
 specifying the document `type` in the request body.
 
+### File upload changes
+
+* API v1 has a file upload endpoint that lets you upload JSON files.
```
cjcenizal, line 176: I think it would be clearer to name the API directly like this:
> * File Upload API v1 supports uploading JSON files.
> * File Upload API v2 removes support for uploading JSON files. Use the Indexing API v2 instead, if you need to index JSON files.

cjcenizal (approved):
Looks good! Just had a few suggestions and questions.

www/docs/api-reference/indexing-apis/file-upload/supported_file_formats.md
pwoznic: It's proto-specific so I'll make it more clear! I'll also add a note about using the Indexing API for v2.

PR #413, opened 2025-02-21
feat(new): Add documentation for the tech preview of Integrating External LLMs

pwoznic:
Added topics for the tech preview create LLM feature including the long-form API Concepts section and Release Notes.

### DIFF
--- a/www/docs/api-reference/llms-apis/create-llm.md
+++ b/www/docs/api-reference/llms-apis/create-llm.md
@@ -38,7 +38,11 @@ request body specifies the following parameters:
 * `description:` Lets you add metadata about the model. 
 * `model:` Specifies the exact model name, such as `gpt-4`, `claude-2`. 
 * `uri:` Contains the API endpoint where to send requests, ensuring that the 
-  model is correctly connected. 
+  model is correctly connected.
+* `auth`: Specifies the authentication configuration for an LLM as bearer token 
+  authentication or custom header-based authentication.
+* `test_model_parameters`: Specifies any additional parameters that are required 
+  for the LLM during the test call.
 
 ### Authentication methods
 
--- a/www/docs/api-reference/llms-apis/create-llm.md
+++ b/www/docs/api-reference/llms-apis/create-llm.md
@@ -1,0 +1,92 @@
+---
+id: create-llm
+title: Create LLM API Definition
+sidebar_label: Create LLM API Definition
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+import {vars} from '@site/static/variables.json';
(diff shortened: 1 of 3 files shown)


www/docs/release-notes.mdx
```diff
@@ -19,6 +19,29 @@ and how these product and documentation changes can benefit your enterprise.
...
+Vectara introduces the tech preview of the Create LLM API, enabling users to 
+integrate and configure external Large Language Models (LLMs) for use with 
+query and chat endpoints. This API enables connectivity with OpenAI-compatible 
+models, including Anthropic Claude, Azure OpenAI, and custom-hosted LLMs.
+
+**Why it matters:** Organizations using multiple LLMs now have a streamlined 
+way to manage model configurations, authentication, and deployment settings 
+through a single API. This update enhances flexibility, allowing users to 
+connect to external LLM providers, define custom authentication methods, and 
+specify model parameters and API endpoints for their specific use cases.
```
omnomagonz, line 35:
```suggestion
**Why it matters:** Organizations using multiple LLMs now have a streamlined 
way to manage model configurations, authentication, and deployment settings 
through a single API. This update enhances flexibility, allowing users to 
connect to external LLM providers, define custom authentication methods, and 
specify model parameters and API endpoints for their specific use cases.
```

Non-blocking: the second half of this makes sense to me, but the first half might be framing this in a way I'm not sure we mean.

Organizations _can_ and probably _will_ use multiple LLMs, but I think this feature is primarily about enabling orgs to bring their own model(s) for any number of reasons.

It might help to frame this as "Organizations need control over the LLMs they use in their AI application development, including the configuration, authentication, and deployment of those LLMs. This update provides that flexibility ..."

omnomagonz (approved):
@pwoznic Some non-blocking suggestions

pwoznic:
Updated after some wordsmithing. :)

www/docs/api-reference/llms-apis/create-llm.md
```diff
@@ -0,0 +1,95 @@
...
+To create a new LLM configuration, send a `POST` request to `/v2/llms`. The 
+request body specifies the following parameters:
+
+* `type:` Set to `openai-compatible` to indicate compatibility with OpenAI-based 
+  APIs such as OpenAI, Anthropic Claude, or Azure OpenAI.
+* `name:` Provides a user-defined label for the LLM configuration.
+* `description:` Lets you add metadata about the model. 
+* `model:` Specifies the exact model name, such as `gpt-4`, `claude-2`. 
+* `uri:` Contains the API endpoint where to send requests, ensuring that the 
+  model is correctly connected.
```
nikhils98, line 41: @pwoznic 

Two more fields available in the request schema:
- `auth`
- `test_model_parameters` (just added in the spec couple of hours ago)

PR #430, opened 2025-03-26
api(new): Add List Table Extractors and Get OAuth Token to the API Reference

pwoznic:
Refreshed the yaml for the API Reference


pranavh4 (approved):
I don't have much context regarding the other change, but the create encoder api changes lgtm

pranavh4 (changes requested):
I just saw the docs preview and had a few suggestions

image

Can we exclude the `output_dimensions` field in the sample request. Its an advanced configuration option and only a few models support it. Also it'd be nice if the example response had values that match the request

```
{
  "id": "enc_1",
  "name": "my-openai-encoder",
  "type": "openai-compatible",
  "output_dimensions": 1536,
  "description": "OpenAI text embedding encoder",
  "default": true,
  "enabled": true
}
```

pwoznic:
> I just saw the docs preview and had a few suggestions
> 
> image
> 
> Can we exclude the `output_dimensions` field in the sample request. Its an advanced configuration option and only a few models support it. Also it'd be nice if the example response had values that match the request
> 
> ```
> {
>   "id": "enc_1",
>   "name": "my-openai-encoder",
>   "type": "openai-compatible",
>   "output_dimensions": 1536,
>   "description": "OpenAI text embedding encoder",
>   "default": true,
>   "enabled": true
> }
> ```

It has to be hidden in the master yaml and then we can rerun the action to regenerate the API reference.

pranavh4 (approved):
Approved. Request encoder changes will be included in the next PR as that's the easiest way to do it

PR #453, opened 2025-05-01
feat(new): Add documentation for the List Hallucination Correctors API

pwoznic:
Adds another endpoint as part of the new Hallucination Correctors APIs. The API Reference will be updated in a separate PR.


abhishekpradhan (approved):

abhishekpradhan (approved):
Note the update endpoint name

PR #468, opened 2025-06-12
feature(new): Add Query-Aware Hallucination Detection and Correction to the Vectara Hallucination Corrector

pwoznic:
Updated VHC topics and Release Notes

### DIFF
--- a/www/docs/api-reference/llms-apis/hallucination_correctors.md
+++ b/www/docs/api-reference/llms-apis/hallucination_correctors.md
@@ -23,7 +23,13 @@ To correct a potentially hallucinated summary, send a `POST` request to
 * `documents`: One or more source documents containing the factual information that 
   the summary should be based on.
 * `text`: The full content of a source document.
-* `model`: The name of the LLM model to use for hallucination correction.
+* `model_name`: The name of the LLM model to use for hallucination correction.
+* `query`: (Optional) The original user query that led to the generated text. 
+  Including this field enables query-aware correction. The model may use the 
+  query to understand expected formatting or context that is not explicit in 
+  the source documents. To use this field, you must also specify `vhc-large-1.0`
+  as the `model_name`.
+
 
 ### Example request
 
@@ -61,10 +67,10 @@ If the input summary is accurate, the `corrected_summary` matches the `original_
 ### Error responses
 
 * **400 Bad Request** – The request body was malformed or contained invalid 
-  parameters.
+  parameters. This can occur if you use `model` instead of the required `model_name` 
+  parameter.
 * **403 Forbidden** – The user does not have permission to perform factual 
   consistency evaluation.
-
 
 ## REST 2.0 URL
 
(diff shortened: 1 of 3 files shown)


www/docs/api-reference/llms-apis/hallucination_correctors.md
```diff
@@ -23,7 +23,13 @@ To correct a potentially hallucinated summary, send a `POST` request to
 * `documents`: One or more source documents containing the factual information that 
   the summary should be based on.
 * `text`: The full content of a source document.
-* `model`: The name of the LLM model to use for hallucination correction.
+* `model_name`: The name of the LLM model to use for hallucination correction.
+* `query`: (Optional) The original user query that led to the generated text. 
+  Including this field enables query-aware correction. The model may use the 
+  query to understand expected formatting or context that is not explicit in 
+  the source documents. To use this field, you must also specify `vhc-large-1.0`
+  as the `model_name`.
```
husseinVectara, line 31: This is not true, you can specify this field with either model, but internally it only actually changes anything if it's vhc large. I don't think we should expose that. I think we say it just works better for vhc large

www/docs/release-notes.mdx
```diff
@@ -16,6 +16,40 @@ experimenting with Retrieval Augmented Generation (RAG), or exploring our
...
+the optional query field to VHC requests enables the model to interpret 
+response formatting better. This helps resolve ambiguities, and attribute 
+facts that are inferred from the query rather than only the source. This 
+results in improved correction accuracy, particularly for instructions like 
+_“Answer True or False,”_ _“List the top 3,”_ or _“Summarize concisely.”_
+
+**What’s new:**
+
+* The `model` parameter has been renamed to `model_name`,
+* Added an optional `query` field for VHC requests. You must use `vhc-large-1.0` as the `model_name`.
```
husseinVectara, line 42: Don't say must use here. It doesn't break anything if they add query using vhc small. Can u find some way to explain that it works better with vhc large while small doesn't really consider it.

husseinVectara (changes requested):
The optional query parameter is always allowed but it will only change results if they use vhc large. I don't think we should expose this though. I feel that we just find a clever way to mention that vhc large optimized to handle queries with the source documents while small is not. Let's not give too much detail on this

www/docs/api-reference/llms-apis/hallucination_correctors.md
```diff
@@ -63,6 +63,31 @@ The response corrects the original summary.
...
+{
+  "original_text": "The Earth has three moons.",
+  "corrected_text": "",
+  "corrections": [
+    {
+      "original_text": "The Earth has three moons.",
+      "corrected_text": "",
+      "explanation": "The source states that the Earth has only one moon. The entire statement is factually incorrect."
+    }
+  ]
```
koala-woman, line 87: I think the example is still a little confusing because in your current example, the corrected text would be "the Earth has one moon"

. Maybe use this one:

"original_text": "According to Martian Guide to Humanity, The Earth has three moons.",
  "corrected_text": "",
  "explanation": "There is no source found for Martian Guide to Humanity (hallucinated source), and there is no source for the earth having three moons. The entire statement is factually incorrect."

koala-woman (approved):

PR #481, opened 2025-07-15
platform(update): Enhance API Reference usability with automated API key type badges

pwoznic:
Provides color-coded badges for each endpoint based on a custom tag in the yaml file. This enhancement helps users understand which API key type is supported for any endpoint. For example, some require the Personal API key but also let you use the QueryService key. Other endpoints might only allow the Personal API key.

On the platform side, there is a new `x-required-api-key-type` tag that will be added to each endpoint in one of the following ways:

x-required-api-key-type:
- Personal

x-required-api-key-type:
- Personal
- QueryService

x-required-api-key-type:
- Personal
- IndexService


www/src/theme/ApiItem/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+      const api = JSON.parse(
+        new TextDecoder().decode(ungzip(base64ToUint8Array(frontMatter.api)))
+      );
+      
+      // Access the custom x-required-api-key-type extension
+      apiKeyTypes = api['x-required-api-key-type'] || null;
+      
+    } catch (error) {
+      console.error('Failed to parse API data:', error);
+      apiKeyTypes = null;
```
cjcenizal, line 40: I don't think this line is needed.

